PR #6, opened 2016-11-06
Packaging fixes

purcell:
- Add autoload cookie
- Remove spurious "ends here" line
- Fix header formatting, and improve commentary
- Fix page-break-lines dependency version

(I feel like `dashboard-insert-startupify-lists` should just be called `dashboard`, and also be autoloaded.)


rakanalh:
Thanks. I'll consider your suggestion with a bit of refactoring later on.

PR #7, opened 2016-11-06
Move page separator to a `Custom` option

wynro:
Simple, I don't think I can say a lot here.


rakanalh:
Yup, simple and beautiful. Thanks alot!

PR #9, opened 2016-11-06
Inhibit emacs default startup screen

andjscott:


rakanalh:
Looks good... thanks alot!

Issue #8, opened 2016-11-06
Allow startup list customization


wynro:
As commented by purcell in melpa/melpa#4393, the startup list is harcoded.

How about giving the user 2 lists, one of pairs `(:symbol . 'function)`, and another of `:symbols`, the first list defines the relationship between symbols (options) and the functions that will create them. The second will define the ordering of the elements. That way each user can define their own personal options easily.

andjscott:
I haven't programmed with elisp so but how does this look? https://github.com/andjscott/emacs-dashboard/blob/customizable/dashboard.el#L224

rakanalh:
Hello @andjscott,

Could you please open an PR? It's easier to discuss changes in the PR page.

rakanalh:
merged into master

PR #10, opened 2016-11-07
[WIP] Allow startup list customization

andjscott:
>[Give] the user 2 lists, one of pairs (:symbol . 'function), and another of :symbols, the first list defines the relationship between symbols (options) and the functions that will create them. The second will define the ordering of the elements. That way each user can define their own personal options easily.

From  https://github.com/rakanalh/emacs-dashboard/issues/8#issuecomment-258765120

### DIFF
--- a/README.org
+++ b/README.org
@@ -37,6 +37,24 @@ git clone https://github.com/rakanalh/emacs-dashboard ~/.emacs.d/dashboard
   (dashboard-setup-startup-hook))
  #+END_SRC
 
+To only display the 5 most recent files and projects:
+#+BEGIN_SRC elisp
+(require 'dashboard)
+(setq dashboard-items '((recents  . 5)
+			(projects . 5)))
+(dashboard-setup-startup-hook)
+ #+END_SRC
+
+To add custom items:
+#+BEGIN_SRC elisp
+(require 'dashboard)
+(defun dashboard-insert-custom ()
+  (insert "Custom text"))
+(add-to-list 'dashboard-item-generators  '(custom . dashboard-insert-custom))
+(add-to-list 'dashboard-items '(custom) t)
+(dashboard-setup-startup-hook)
+ #+END_SRC
+
 * Shortcuts
 
 You can use any of the following shortcuts inside Dashboard
--- a/dashboard.el
+++ b/dashboard.el
@@ -22,8 +22,15 @@
 ;;; Code:
 
 (eval-when-compile (require 'bookmark))
-(eval-when-compile (require 'projectile))
 (eval-when-compile (require 'page-break-lines))
+
+(defun dashboard-subseq (seq start end)
+  "Use `cl-subseq`, but accounting for end points greater than the size of the
+list. Return entire list if `END' is omitted."
+  (let ((len (length seq)))
+    (cl-subseq seq start (and (number-or-marker-p end)
+                              (min len end)))))
+
 
 ;; Custom splash screen
 (defvar dashboard-mode-map
@@ -70,35 +77,11 @@
 
 (defun dashboard-insert-banner ()
   "Insert Banner at the top of the dashboard."
+  (goto-char (point-max))
   (dashboard-insert-ascii-banner-centered
    (concat (file-name-directory
 	    (locate-library "dashboard"))
 	   "banner.txt")))
-
-(defun dashboard-insert-startupify-lists ()
-  "Insert the list of widgets into the buffer."
-  (interactive)
-  (with-current-buffer (get-buffer-create "*dashboard*")
-    (let ((buffer-read-only nil)
-          (list-separator "\n\n"))
-      (goto-char (point-max))
-      (dashboard-insert-banner)
-      (dashboard-insert-page-break)
-
-      (recentf-mode)
-      (when (dashboard-insert-file-list "Recent Files:" (recentf-elements 5))
-	(dashboard-insert--shortcut "r" "Recent Files:")
-	(dashboard-insert-page-break))
-
-      (when (dashboard-insert-bookmark-list "Bookmarks:" (bookmark-all-names))
-	(dashboard-insert--shortcut "m" "Bookmarks:")
-	(dashboard-insert-page-break))
-
-      (projectile-mode)
-      (when (dashboard-insert-project-list "Projects:" (projectile-relevant-known-projects))
-	(dashboard-insert--shortcut "p" "Projects:")
-	(dashboard-insert-page-break)))
-    (dashboard-mode)))
 
 (defun dashboard-insert-file-list (list-display-name list)
   "Render LIST-DISPLAY-NAME title and items of LIST."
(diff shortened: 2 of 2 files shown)


dashboard.el
```diff
@@ -180,6 +163,78 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
     (beginning-of-line)
     (widget-forward 1)))
 
+(defun dashboard-generate-banner ()
```
rakanalh: Banner is not a selectable widget, it should always be there so i would remove it from the list of widgets

dashboard.el
```diff
@@ -180,6 +163,78 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
+  (when (goto-char (point-max))
+    (dashboard-insert-banner)
+    (dashboard-insert-page-break)))
+(defun dashboard-generate-recents ()
```
rakanalh: Would be nice to have an empty line before each `defun`

dashboard.el
```diff
@@ -180,6 +163,78 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
+    (dashboard-insert-page-break)))
+(defun dashboard-generate-bookmarks ()
+  (require 'bookmark)
+  (when (dashboard-insert-bookmark-list
+	 "Bookmarks:"
+	 (dashboard-subseq (bookmark-all-names)
+			   0 list-size))
+    (dashboard-insert--shortcut "m" "Bookmarks:")
+    (dashboard-insert-page-break)))
+(defun dashboard-generate-projects ()
```
rakanalh: To be honest, i don't like the `generate` name. Since the call is `dashboard-insert-startupify-list` i would go for `insert`

dashboard.el
```diff
@@ -180,6 +163,78 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
+	 (dashboard-subseq (projectile-relevant-known-projects)
+			   0 list-size))
+    (dashboard-insert--shortcut "p" "Projects:")
+    (dashboard-insert-page-break)))
+
+(defun dashboard-generate-fallback ()
+  (insert "Undefined dashboard item")
+  (dashboard-insert-page-break))
+
+(defvar dashboard-item-generators  '((banner    . dashboard-generate-banner)
```
rakanalh: Banner should always be inserted

dashboard.el
```diff
@@ -180,6 +163,78 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
+				     (recents   . dashboard-generate-recents)
+				     (bookmarks . dashboard-generate-bookmarks)
+				     (projects  . dashboard-generate-projects)))
+(defvar dashboard-items '((banner)
```
rakanalh: Banner should always be inserted

dashboard.el
```diff
@@ -180,6 +163,78 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
+			  (recents   . 5)
+			  (bookmarks . 5)
+			  (projects  . 7))
+  "Association list of items to show in the startup buffer of the form
+`(list-type . list-size)`. If nil it is disabled.
+Possible values for list-type are:
+`banner' `recents' `bookmarks' `projects'
+")
+(defvar dashboard-items-default-length 20
```
rakanalh: Let's also move all defvars to the top after the mode declaration.

dashboard.el
```diff
@@ -180,6 +163,78 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
+  (require 'projectile)
+  (projectile-mode)
+  (when (dashboard-insert-file-list
+	 "Projects:"
+	 (dashboard-subseq (projectile-relevant-known-projects)
+			   0 list-size))
+    (dashboard-insert--shortcut "p" "Projects:")
+    (dashboard-insert-page-break)))
+
+(defun dashboard-generate-fallback ()
```
rakanalh: I dont like this fallback. Can we just do nothing if we encounter an invalid widget? or make emacs blow up and the user will figure out what is wrong

rakanalh:
@wynro care to provide feedback on your side as well?

rakanalh:
@andjscott Thanks alot, appreciate your help to get this done 👍  have a look at my feedback please

dashboard.el
andjscott: Could `banner` be prepended to the list of items like [spacemacs does for warnings](https://github.com/syl20bnr/spacemacs/blob/b7e51d70aa3fb81df2da6dc16d9652a002ba5e6b/core/core-spacemacs-buffer.el#L881)?

rakanalh: They don't treat the banner as a widget https://github.com/syl20bnr/spacemacs/blob/b7e51d70aa3fb81df2da6dc16d9652a002ba5e6b/core/core-spacemacs-buffer.el#L827 and neither will we. The warnings in this case IS a widget that displays some list but the banner is not.

README.org
```diff
@@ -37,6 +37,16 @@ git clone https://github.com/rakanalh/emacs-dashboard ~/.emacs.d/dashboard
   (dashboard-setup-startup-hook))
  #+END_SRC
 
+To add custom items
```
rakanalh: This actually wasn't what i was referring to. What i wanted you to add is a documentation on how to customize the existing list rather than creating a new item.

So for example, the current default shows everything, but i only want to show bookmarks or (bookmarks + recent files)... how do i do that? we should have this documented using your approach.

dashboard.el
```diff
@@ -25,6 +25,14 @@
 (eval-when-compile (require 'projectile))
```
rakanalh: This could be removed btw given that projectile is only needed if we have the projectile widget in the list

andjscott: Could `(eval-when-compile (require 'bookmark))` go too?

Does the require at [dashboard.el#L183](https://github.com/andjscott/emacs-dashboard/blob/03c169f3849e691b58101a2c7b7363d8f2ea6f70/dashboard.el#L183) need to be changed to `(eval-when-compile (require 'projectile))`?

rakanalh:
@andjscott Wheew! that was a thorough review? Can you now squash your commits into 1 which summarizes the change?

rakanalh:
Ok great. I checked out your branch and added the configs you provide as an example in the README:
```
(require 'dashboard)
(setq dashboard-items '((recents  . 5)
			(projects . 5)))
```

It still shows bookmarks which should be skipped right?

andjscott:
Um works-on-my-machine™ with a pretty minimal init.el

rakanalh:
Not for projects on my machine at least... i'll have a thorough look.

Thanks!

andjscott:
I'm not sure why it could be. Do bookmarks still show if you delete the`(bookmarks . 5)` line from `dashboard-items` and all the customization from init.el?

wynro:
In my machine works also perfectly, maybe you have compiled it and emacs is taking the compiled copy instead of the correct one?

rakanalh:
@wynro i recompiled it but it doesnt seem to pick up projects... for recentf it works but i get an empty block for projects. While if i remove the `setq`, it works with all 3 components.

rakanalh:
It's weird. Seems like `projectile-relevant-known-projects` is returning an empty list at first. But then when i execute the function manually, the list is returned.
The other weird behaviour is that, recentf + projects doesn't work while bookmarks + projects works.

PR #14, opened 2016-11-20
Ensures *dashboard* buffer exists at startup

### LINKED ISSUE #13: Initial startup always asks to create buffer
I've worked around this by pre-creating the `*dashboard*` buffer in my startup definition:
```lisp
(use-package dashboard
  :config
  (get-buffer-create "*dashboard*")
```
In the emacs startup sequence, the startup hook is [invoked later][1]  than the init hook, 
so perhaps the order should switch the around?
(i.e switch-to-buffer on startup, do work in init hook?)

[1]: http://emacs.stackexchange.com/questions/14551/whats-the-difference-between-after-init-hook-and-emacs-startup-hook

mgrbyte:
Fixes #13 
* Adds content to buffer in `after-init-hook`
* Switches to buffer, enables `page-break-line-mode` and rendering in `emacs-startup-hook` 
* Adds Cask build system for developing packages (see individual commit messages below)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 
 # Packaging
 .cask
+/dist/dashboard-1.0.0.tar
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,5 @@
 
 # Packaging
 .cask
-/dist/dashboard-1.0.0.tar
+/dist
+
--- a/Cask
+++ b/Cask
@@ -1,0 +1,8 @@
+(source gnu)
+(source melpa)
+
+(package-file "dashboard.el")
+(files "*.el" "*.org" "*.txt")
+				
+(development
+ (depends-on "page-break-lines"))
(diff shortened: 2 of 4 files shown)


rakanalh:
Hey @mgrbyte 

Thank you for considering fixing your own issue :)

One question, why did you add `dashboard-pkg.el`? does it provide meta information to something i don't know about?

I am asking because i am not sure i like maintaining the sync between `dashboard.el` which already provides version & dependencies... along with the Caskfile and now `dashboard-pkg.el`. 

If it's not an absolute must, i just keep the current package without that and preferably without the caskfile as well.

mgrbyte:
@rakanalh It's not an absolute must, it's just that I use cask, and it was the quickest way for me to develop and test this fix.
I'd be interested to know how you package this and deploy it to Melpa (there are other methods).
cask is just one way to help automate that.

mgrbyte:
When creating emacs-lisp packages, a -pkg.el file is [required][1] in order for the installation to work correctly with the package.el system.
I'm guessing that the tool/command you use to build the package and deploy to melpa generates one for you automatically.

[1]: https://www.gnu.org/software/emacs/manual/html_node/elisp/Multi_002dfile-Packages.html#Multi_002dfile-Packages

rakanalh:
I wasn't aware -pkg.el file should also be present. I thought the docs at the top of the file should be enough for MELPA & package.el system. 

Given this, looks good to me... thanks alot!

PR #17, opened 2016-11-20
Document dev-builds and provide simple Makefile.

mgrbyte:
@rakanalh Having introduced the -pkg.el file as discussed in comments on #14,
I've realised that the -pkg.el file does _not_ need to maintained, since the build will generate that file and place it in the tar file (in the `build` directory as per this PR).

 * Avoid repeating version and dependency information.


rakanalh:
To be honest, i tested this PR with Cask but i don't like it's approach. I think it's a bit much for me at this level. I'll stick to `package-install-file` for now.

Thank you for the help... much appreciated.

mgrbyte:
@rakanalh fair enough! I am a bit confused though, as using this PR, using `package-install-file` is still necessary ... the only differences (to current master) are removal of the `dashboard-pkg.el` file
and addition of the Makefile (to serve as both documentation for using `cask`, and to automate its usage).

Anyway, glad if what I've contributed helped.

rakanalh:
Cask confused me a bit to be honest.

1. The check `if [ -d .cask ]` didn't work because .cask did not exist. `[ -f Cask ]` worked. 
2. Also `@cask package ${BUILD_DIR} --dev` didn't do anything in my case, for some reason the package wasn't being installed into my .emacs/elpa which is what i expected. And there doesn't seem to be anything documented about this so i wonder where this repo should be placed in the first place. In my case it's nowhere near .emacs.d.

So kinda got confused by Cask that's why i just decided to keep it simple.

You've definitely helped alot. Thanks alot and i hope this doesn't stop you from contributing further awesome stuff.

mgrbyte:
@rakanalh Ah OK, I see;
So it looks like I messed up the Makefile (I should of just kept it simple, and let it with a simple instructions in the README.org).

FWIW:
`cask init --dev`: 
   creates a .cask directory in the current directory (`~/git/emacs-dashboard` in my case (the location I cloned my fork of this repo to)

`cask package --dev build`
  Reads the `Cask` file and from that, determines the files to include in the package-archive stored in `build/emacs-dashnoard-<version>.tar`

To

PR #23, opened 2016-12-04
Dont show dashboard when command line arguments are present

qorrect:
This also works for any flag, for example emacs --debug-init , you probably dont want dashboard showing either ?

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -96,6 +96,7 @@ Set to nil for unbounded.")
        (while (not (eobp))
          (let ((line-length (- (line-end-position) (line-beginning-position))))
            (if (< banner-width line-length)
+
                (setq banner-width line-length)))
          (forward-line 1))
        (goto-char 0)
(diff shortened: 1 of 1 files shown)


dashboard.el
```diff
@@ -96,6 +96,7 @@ Set to nil for unbounded.")
        (while (not (eobp))
          (let ((line-length (- (line-end-position) (line-beginning-position))))
            (if (< banner-width line-length)
+
```
rakanalh: Could you please undo this?

dashboard.el
```diff
@@ -256,15 +257,18 @@ Optionally, provide NO-NEXT-LINE to move the cursor forward a line."
...
-
-  (add-hook 'emacs-startup-hook '(lambda ()
-				   (switch-to-buffer "*dashboard*")
-				   (goto-char (point-min))
-				   (redisplay))))
+
+  ;; If we have command line arguments, we just assume filenames and dont display the Dashboard
+  (if (< (length command-line-args) 2 )
+      (progn
+	(setq inhibit-startup-screen t)
```
rakanalh: We already do this so no need to re-add it

qorrect:
I committed to my forks master branch, will that work or do I need to do something else ?

rakanalh:
Yup... i merged it into a branch so that i can add a couple of things.

Thanks

PR #25, opened 2016-12-05
Add reddit /r/emacs section to dashboard 

qorrect:
Should only make the request if reddits is explicitly set to > 0 .

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -75,14 +75,15 @@ Return entire list if `END' is omitted."
 
 (defvar dashboard-item-generators  '((recents   . dashboard-insert-recents)
                                      (bookmarks . dashboard-insert-bookmarks)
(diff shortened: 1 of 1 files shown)


rakanalh:
This looks pretty awesome... i would definitely use in my emacs setup.

However, that's the whole point of providing custom dashboard items. It is basically there so that others can provide functions that can be hooked into dashboard and used. But i disagree that this widget has a place in the official dashboard package. You can provide it as a mepla package to be used by people who want to do so?

Many thanks!

qorrect:
Ok will do, looking up the instructions for adding it to melpa, it will eventually be here https://github.com/qorrect/emacs-dashboard-reddits.

Going to whip one up for emacs.stackexchange.com as well, and probably news.ycombinator too , why not.  Thanks for all your help!

PR #41, opened 2017-04-19
Fix agenda due-date issues

eigengrau:
This fixes two issues when populating the list of today’s due agenda entries.


rakanalh:
Hello @eigengrau 

Thank you for all those nice PRs.

I am inclined to merge this PR but not including the last commit. I don't really like the fact that we should hide "stale" items because user's requirements may vary. I think i would rather have that item nag me until i actually get to do it rather than have my tools hide it for me.

eigengrau:
Hello @rakanalh. Makes sense. I briefly thought the original org-agenda would also hide stale deadlines, but it doesn’t.

PR #43, opened 2017-04-19
org-agenda: include today’s appointments

eigengrau:
NB: to avoid merge conflicts I’ve based these changes on my previous commits (cf. other PRs).


eigengrau:
Unfortunately, org does not expose any proper API to handle date ranges, so these are not supported yet. The only way I can see so far is doing things mostly manually, by using `org-element-timestamp-parser` (takes no parameters, works only on thing-at-point).

JesusMtnez:
@eigengrau is this change still necessary?

JesusMtnez:
This issue is not happening right now. Please feel free to open a new one if necessary.

PR #53, opened 2017-09-23
Add register widget

pedros:
This patch adds a new widget for displaying register lists. It uses `register-describe-oneline` to format each register. Its output might look like this: 

    Registers:
        z - file "~/.zshrc".
        X - file "~/.Xresources".
        x - file "~/.xinitrc".
        t - file "~/.tmux.conf".
        r - file "~/.ratpoisonrc".

The above only shows files, but the widget supports any register type, and will format it just as `register-list` would.


rakanalh:
Very nice! thank you very much

PR #59, opened 2017-12-27
Customizability improvements

TonCherAmi:
Hi, this PR adds faces for widget headings and the banner title which allows for this kind of stuff:

screen

Both of those faces are set to inherit from the `default` face by default so the initial look will remain the same.

It also adds an option to disable the banner title altogether by setting `dashboard-banner-logo-title` to `nil`.

Please do not hesitate to point out any mistakes I may have made or make any suggestions of your own!


rakanalh:
That's pretty awesome! Thank you for taking the time to do this.

I will test it during the upcoming weekend and merge.

rakanalh:
Many thanks!

PR #60, opened 2018-01-07
load the banner image via imagemagick when available

gonewest818:
This PR causes banner images to be loaded with ImageMagick when Emacs is compiled `--with-imagemagick`.   Doing so extends support for additional file formats, and makes it possible to optionally resize the image while loading.

I needed this because my Emacs config grabs an "image of the day" from an external site and I want that image to be my dashboad banner, but I want to be flexible about the file formats and resolutions I might encounter in the wild.

Separately, I expect to create an "image of the day" widget as well.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -16,9 +16,28 @@
 ;;
 ;; Customs
 ;;
+
 (defcustom dashboard-page-separator "\n\f\n"
   "Separator to use between the different pages."
   :type 'string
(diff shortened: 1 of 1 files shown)


rakanalh:
Hello @gonewest818,

Thank you so much for this PR, it looks great. I'll test it tonight or tomorrow and merge if all is good.

rakanalh:
Many Thanks again.

PR #68, opened 2018-05-19
Fix dashboard-setup-startup-hook

Indy2222:
Hooks from `emacs-startup-hook` are called too early so dashboard
incorrectly drawn text version of logo on some occasions. Steps to
reproduce:

* Start `emacs --daemon` from terminal
* Open `emacsclient -t`
* Open `emacsclient -c` -- text version of logo is displayed

This behavior is fixed by utilizing `window-setup-hook`. Emacs'
documentation of `window-setup-hook`:

This normal hook is very similar to emacs-startup-hook. The only
difference is that it runs slightly later, after setting of the frame
parameters.

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -155,10 +155,10 @@ If a command line argument is provided, assume a filename and skip displaying Da
 	(add-hook 'after-init-hook (lambda ()
 				     ;; Display useful lists of items
 				     (dashboard-insert-startupify-lists)))
-	(add-hook 'emacs-startup-hook '(lambda ()
-					 (switch-to-buffer "*dashboard*")
-					 (goto-char (point-min))
-					 (redisplay))))))
+	(add-hook 'window-setup-hook '(lambda ()
+                                        (switch-to-buffer "*dashboard*")
+                                        (goto-char (point-min))
+                                        (redisplay))))))
 
 (provide 'dashboard)
 ;;; dashboard.el ends here


winny-:
Thanks, I had noticed this too, but hadn't found time to look into it.

Indy2222:
But it looks like this doesn't fix the problem or at least not in all cases. I think that the correct solution would be to call `dashboard-insert-startupify-lists` after a new frame is created if `*dashboard*` buffer is selected or something like that.

rakanalh:
Hello @Indy2222,

I haven't tested this branch yet,,, but seeing your last commit, you think that this PR does not solve the issue you guys are facing with dasboard on emacsclient, correct?

Indy2222:
> I haven't tested this branch yet,,, but seeing your last commit, you think that this PR does not solve the issue you guys are facing with dasboard on emacsclient, correct?

You are right, it seems that the patch doesn't fix the issue. Unfortunately, I haven't had time yet to change it. It seems that calling `dashboard-insert-startupify-lists` fixes the problem and judging on `dashboard-resize-on-hook` is the right thing to do.

`(add-hook 'window-configuration-change-hook 'dashboard-resize-on-hook)` ensures that the buffer is properly displayed after changes to its window (if I get it right), IMHO we should call it (or a small modification of it) after new frame is created.

I will appreciate any feedback or help with this as my experience with Elisp is close to zero.

JesusMtnez:
> Hooks from `emacs-startup-hook` are called too early so dashboard
> incorrectly drawn text version of logo on some occasions. Steps to
> reproduce:
>     * Start `emacs --daemon` from terminal
>     * Open `emacsclient -t`
>     * Open `emacsclient -c` -- text version of logo is displayed

@Indy2222 Are you fancy this problem nowadays? I've tried to reproduce it but it worked well :thinking:

Indy2222:
> @Indy2222 Are you fancy this problem nowadays? I've tried to reproduce it but it worked well thinking

I haven't encountered the issue for some time (Emacs 26.2).

Zeta611:
Hi, I'm suffering from this issue (using version 20220615.540) in Emacs 28.1.

When I start emacs via `emacs --daemon` and attach to it with `emacsclient -c`, the logo and the file icons do not load.
As soon as I change the size of the window, it immediately shows the logo and the icons, and subsequent `emacsclient -c`s display them without resizing the windows.

The relevant part of my config is:
```emacs-lisp
(use-package dashboard
    :ensure t
    :config
    ; show dashboard after emacsclient -c
    (setq initial-buffer-choice
      (lambda () (get-buffer-create "*dashboard*")))

    (setq dashboard-banner-logo-title "Hi, Jay!")
    (setq dashboard-setup-banner 'official)
    (setq dashboard-set-heading-icons t)
    (setq dashboard-set-file-icons t))
```

Issue #71, opened 2018-05-31
window-configuration-change-hook is deprecated


vxid:
The `window-configuration-change-hook` is deprecated with Emacs 26. According to the changelog of v26:

> Resizing a frame no longer runs 'window-configuration-change-hook'. 'window-size-change-functions' should be used instead.

For dashboard, `(dashboard-resize-on-hook)` is not called when the Emacs window is resized by my window manager, resulting in a off-center image. Changing the frame configuration recenters it though.

PR #72, opened 2018-05-31
Fix deprecated window-configuraton-change-hook

vxid:
See #71 

My skills in Elisp are pretty limited but I get a normal resizing behaviour after switching to my fork.


rakanalh:
Hi @vxid,

Thanks! the change is interesting. I'll test it during next week.

vxid:
Thanks!
Note that I amended my commit since the method used in justbur/emacs-which-key@0dc4e84 seems cleaner.

PR #75, opened 2018-07-14
Update README to include `:ensure` keyword in `use-package` instructions

mveytsman:
This will automatically install the package.

### DIFF
--- a/README.org
+++ b/README.org
@@ -30,6 +30,7 @@ M-x package-install RET dashboard
 (dashboard-setup-startup-hook)
 ;; Or if you use use-package
 (use-package dashboard
+  :ensure t
   :config
   (dashboard-setup-startup-hook))
  #+END_SRC


rakanalh:
Hi @mveytsman 
As far as i know, use-package will download and automatically install the package without the :ensure flag. At least that's what i have in my setup.
Could you describe why you needed this flag in your setup?

mveytsman:
@rakanalh In my set up I get the following error: `Error (use-package): Cannot load dashboard` 

My understanding of the [docs](https://github.com/jwiegley/use-package#package-installation) is that the ensure keyword is what makes use-package download the package if it's not present.

rakanalh:
I see. I thought that use-package does that automatically but i checked my configs and found that i was using `(setq use-package-always-ensure t)`. So your change makes sense.

Thanks @mveytsman

PR #76, opened 2018-07-22
Fixed typo

purpleGreap:
On line 64, added a matching `"`  since single `"` isn't enough to load a png file (resulting in error).

### DIFF
--- a/README.org
+++ b/README.org
@@ -61,7 +61,7 @@ To update the banner or banner title
 ;; 'official which displays the official emacs logo
 ;; 'logo which displays an alternative emacs logo
 ;; 1, 2 or 3 which displays one of the text banners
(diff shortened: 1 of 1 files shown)


rakanalh:
Hi @purpleGreap, this is a docstring for the function and fixing the terminating quote there won't resolve any errors. Speaking of which, are you facing any errors with Dashboard?

purpleGreap:
Hello, I am a new emacs user so I followed the instruction and wrote `(setq dashboard-startup-banner "/path/to/image)` which resulted in error so I used this code instead `(setq dashboard-startup-banner "/path/to/image")` and it worked. Hence I thought that it was a typo. Please let me know if I'm wrong. And also thank you for this awesome package. Genuinely helps a lot ! :+1:

PR #79, opened 2018-09-02
Fix dashboard crashing when bookmark isn't a file.

nnoodle:
This fixes dashboard crashing when it tries to insert a bookmark that does not have a filename parameter (e.g. elfeed bookmarks).


rakanalh:
Hi @nnoodle,

It works for me when the bookmarks are either files or directories... out of curiosity, what type of bookmarks did you have that caused the failure this PR is fixing?

nnoodle:
The type of bookmarks that caused it to crash for me are elfeed filters.

rakanalh:
Cool, thank you!

PR #83, opened 2019-01-02
Don't add org-agenda, etc to the recentf list when dashboard renders

MikeTheGreat:
When dashboard starts up it generates the org-agenda, which it does by opening a bunch of my .org files, which are then pushed to the top of the recentf list.

This change will prevent that by replacing the recentf-list with a copy of the first N items in the recentf-list (so that the recentf widget can display them in the dashboard) and then restoring the original list afterwards (thus discarding the new files that were incidentally added as part of dashboard's rendering)

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -1,4 +1,3 @@
-
 ;;; dashboard.el --- A startup screen extracted from Spacemacs
 
 ;; Copyright (c) 2016 Rakan Al-Hneiti & Contributors
--- a/dashboard.el
+++ b/dashboard.el
@@ -1,3 +1,4 @@
+
 ;;; dashboard.el --- A startup screen extracted from Spacemacs
 
 ;; Copyright (c) 2016 Rakan Al-Hneiti & Contributors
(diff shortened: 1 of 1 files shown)


rakanalh:
Hello @MikeTheGreat,

I have the following snippet set in my configuration to exclude those files from my recentf list:

```
(setq recentf-exclude '("/elpa/" ;; ignore all files in elpa directory
                          ".*?autoloads.el$"
                          "/tmp/" ;; ignore temporary files
                          "*/.elfeed/index"
                          "company-statistics-cache.el"
                          ".gitignore"
                          "*/Documents/org-mode"
                          )
```

Therefore, i don't know if this PR is really necessary as you could tell recentf to just ignore these files in the first place.

ferdy:
@rakanalh From my perspective, this PR makes a lot of sense to me.

In my org-directory I have a lot of files automatically generated (weather.org, wta.org, soccer.org, birthday.org and so on) which I never edit them directly, thus I wouldn't like them showing up in my dashboard.

On the other hand, there are several files (e.g. projects files, blog post, etc. etc) which I edit within emacs and I would fine handly to have them showing up in my dashboard.

Just my two cents and thank you very much for the awesome package! Cheers!
--
ferdy

dashboard.el
```diff
@@ -1,3 +1,4 @@
+
```
rakanalh: Please remove this empty line.

MikeTheGreat:
Done!
(I made the change on my branch - that shows up here automatically, yes?)

rakanalh:
@MikeTheGreat did you test this? i did just now and i still have the org files showing up.

MikeTheGreat:
It worked on my machine :)

I did test it (and use it) and it seemed to work fine for me.  

Is something else opening those files up?  Maybe running the Org Agenda command outside of / after dashboard?

In order to see this I did need to start emacs (which runs dashboard for me), then open another file (which adds it to recentf-list), and then quit.  At that point when you restart you'll see that the new file is at the top of the list.  (The org files are still there, but they're not bumping everything else down).

rakanalh:
Ah ok,,, i now see your test case working for me as well.

MikeTheGreat:
That confused me too the first time I ran my code :)

Thanks for accepting this PR!  I'm glad to help contribute!

Thank you again for making dashboard & sharing it!

PR #85, opened 2019-01-26
Add ability to set dashboard-banner-logo-title to a function.

functionreturnfunction:
Simple callback functionality for the logo title; I had the idea when I read [this post](https://www.reddit.com/r/emacs/comments/8jaflq/tip_how_to_use_your_dashboard_properly/) and decided against using hooks because one of my custom dashboard items is a little expensive to run so I don't want to run it twice.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -57,7 +57,7 @@ to the specified width, with aspect ratio preserved."
   "Width of a banner.")
 
 (defvar dashboard-banner-logo-title "Welcome to Emacs!"
-   "Specify the startup banner.")
+   "Specify the startup banner.  If set to a function, that function must take no arguments and return a string.")
 
 (defvar dashboard-startup-banner 'official
    "Specify the startup banner.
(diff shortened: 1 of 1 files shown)


rakanalh:
Hey @functionreturnfunction, could you show me a use-case for this? i would go for hooks being an idiom in the emacs community vs having callbacks for values that only set a string.

functionreturnfunction:
I'm using it to display start time/GC stats, and subsequently last refresh time, like so:

```
(let ((called 0))
    (defun my-dashboard-banner ()
      "Return a string including current date/time as well as information on
startup time and garbage collections.  Subsequent calls only return a string
containing the current date/time."
      (if (< 1 called)
          (format "Last refreshed at %s" (current-time-string))
        (setq called (1+ called))
        (format "Emacs ready at %s in %.2f seconds with %d garbage collections."
                (current-time-string)
                (float-time (time-subtract after-init-time before-init-time))
                gcs-done))))

(setq dashboard-banner-logo-title #'my-dashboard-banner)
```
`after-init-time' can't be derived before dashboard is initialized, which is why OP in the post I linked adds `dashboard-refresh-buffer' to `after-init-hook'.  That would be a bit expensive for me because I have some custom widgets.

belak:
I really wanted a way to specify just a title, rather than a PNG logo. It looks like this would support that... what would need to be done to get this in?

PR #88, opened 2019-02-21
Feature/center content

jkdufair:
Hello @rakanalh. This PR addresses https://github.com/emacs-dashboard/emacs-dashboard/issues/77  
I approached this by creating a macro that can insert arbtrary sections and returns the maximum line length for that section. Then `'dashboard-insert-startupify-lists` finds the maximum of the maximums and moves all items over accordingly. I initially looked at adjusting the margins and/or the gutter but neither looked very attractive.

Additionally, I'd be willing to be a maintainer. I've been looking for something fun to hack on for a while. Thanks for considering.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -287,7 +287,7 @@ WIDGET-PARAMS are passed to the \"widget-create\" function."
    recentf-list
    list-size
    "r"
-   (lambda (&rest ignore) (find-file-existing el))
+   `(lambda (&rest ignore) (find-file-existing ,el))
    (abbreviate-file-name el)))
 
 ;;
@@ -302,7 +302,7 @@ WIDGET-PARAMS are passed to the \"widget-create\" function."
                      0 list-size)
    list-size
    "m"
-   (lambda (&rest ignore) (bookmark-jump el))
+   `(lambda (&rest ignore) (bookmark-jump ,el))
    (let ((file (bookmark-get-filename el)))
      (if file
          (format "%s - %s" el (abbreviate-file-name file))
@@ -323,7 +323,7 @@ WIDGET-PARAMS are passed to the \"widget-create\" function."
                            0 list-size)
          list-size
          "p"
-         (lambda (&rest ignore) (projectile-switch-project-by-name el))
+         `(lambda (&rest ignore) (projectile-switch-project-by-name ,el))
          (abbreviate-file-name el)))))
 
 ;;
@@ -396,11 +396,11 @@ date part is considered."
      (or agenda '())
      list-size
      "a"
-     (lambda (&rest ignore)
-       (let ((buffer (find-file-other-window (nth 4 el))))
-         (with-current-buffer buffer
-           (goto-char (nth 3 el)))
-         (switch-to-buffer buffer)))
+     `(lambda (&rest ignore)
+        (let ((buffer (find-file-other-window (nth 4 ,el))))
+          (with-current-buffer buffer
+            (goto-char (nth 3 ,el)))
+          (switch-to-buffer buffer)))
      (format "%s" (nth 0 el)))
     (and (not agenda)
          (insert "\n    --- No items ---"))))
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -227,34 +227,35 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
 ;;
 ;; Section insertion
 ;;
-(defmacro dashboard-insert-section-list (list-display-name list action &rest rest)
-	"Insert a LIST of items with LIST-DISPLAY-NAME, expanding ACTION and passing REST to widget creation."
+(defmacro dashboard-insert-section-list (section-name list action &rest rest)
+	"Insert a LIST of items with SECTION-NAME, expanding ACTION and passing REST to widget creation."
 	`(let ((max-line-length 0))
-		(when (car ,list)
-			(dashboard-insert-heading ,list-display-name)
-			(mapc (lambda (el)
-							(let ((widget nil))
-								(insert "\n    ")
-								(setq widget
-											(widget-create 'push-button
-																		 :action ,action
-																		 :mouse-face 'highlight
-																		 :follow-link "\C-m"
-																		 :button-prefix ""
-																		 :button-suffix ""
-																		 :format "%[%t%]"
-																		 ,@rest))
-								(setq max-line-length
-											(max max-line-length (length (widget-value-value-get widget))))))
-						,list))
-		max-line-length))
+		 (when (car ,list)
+			 (mapc (lambda (el)
+							 (let ((widget nil))
+								 (insert "\n    ")
+								 (setq widget
+											 (widget-create 'push-button
+																			:action ,action
+																			:mouse-face 'highlight
+																			:follow-link "\C-m"
+																			:button-prefix ""
+																			:button-suffix ""
+																			:format "%[%t%]"
+																			,@rest))
+								 (setq max-line-length
+											 (max max-line-length (length (widget-value-value-get widget))))))
+						 ,list))
+		 max-line-length))
 
 (defmacro dashboard-insert-section (section-name list list-size shortcut action &rest widget-params)
   "Add a section with SECTION-NAME and LIST of LIST-SIZE items to the dashboard.
 ACTION is theaction taken when the user activates the widget button.
 SHORTCUT is the keyboard shortcut used to access the section.
-WIDGET-PARAMS are passed to the \"widget-create\" function."
+WIDGET-PARAMS are passed to the \"widget-create\" function.
+Show EMPTY-LIST-TEXT if no items in list"
 	`(progn
+		 (dashboard-insert-heading ,section-name)
 		 (when-let ((max-line-length
 								 (dashboard-insert-section-list
 									,section-name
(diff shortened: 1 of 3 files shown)


jkdufair:
Just realized I need to update the readme. Sorry about that.

jkdufair:
- Page break lines are now not "centered"
 - Documentation added about centering

rakanalh:
Hello @jkdufair,

Thank you so much for putting together a PR.

Before i review the code, please fix back the code indentation to be able to properly review the other parts of the code. I think we should introduce some lint to that because every contributor seems to have their own indentation style / settings.

jkdufair:
Sure thing. Do you have a style guide I can refer to?

rakanalh:
Not really... now i am considering it to be honest.

jkdufair:
Ok. I tried not to make any whitespace-only changes to existing functions. I'll revert any that I may have created inadvertently. Thank you.

jkdufair:
Ok sorry about that! I didn't realize I had those whitespace-only changes. Thanks for taking a look.

dashboard.el
```diff
@@ -128,20 +133,28 @@
...
+												 (list-size
                           (or (cdr-safe els)
                               dashboard-items-default-length))
-			 (item-generator
+												 (item-generator
                           (cdr-safe (assoc el dashboard-item-generators))))
                     (add-to-list 'dashboard--section-starts (point))
-                    (funcall item-generator list-size)
+                    (setq max-line-length
+													(max max-line-length (or (funcall item-generator list-size) 0)))
```
rakanalh: Indentation is wrong here

rakanalh: Indentation is wrong here and the lines below

dashboard.el
```diff
@@ -128,20 +133,28 @@
...
                     (dashboard-insert-page-break)))
-		dashboard-items))
-	(dashboard-mode)
-	(goto-char (point-min))))
+								dashboard-items)
+					(if dashboard-center-content
```
rakanalh: Please use `(when` for this instead if `(if (progn)`

dashboard-widgets.el
```diff
@@ -225,128 +225,100 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
-                           :button-suffix ""
-                           :format "%[%t%]"
-                           (abbreviate-file-name el)))
-          list)))
+(defmacro dashboard-insert-section-list (section-name list action &rest rest)
+	"Insert a LIST of items with SECTION-NAME, expanding ACTION and passing REST to widget creation."
+	`(let ((max-line-length 0))
+		 (when (car ,list)
+			 (mapc (lambda (el)
+							 (let ((widget nil))
```
rakanalh: Indentation is wrong here and the lines below

dashboard-widgets.el
```diff
@@ -225,128 +225,100 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
+																			:button-prefix ""
+																			:button-suffix ""
+																			:format "%[%t%]"
+																			,@rest))
+								 (setq max-line-length
+											 (max max-line-length (length (widget-value-value-get widget))))))
+						 ,list))
+		 max-line-length))
+
+(defmacro dashboard-insert-section (section-name list list-size shortcut action &rest widget-params)
```
rakanalh: Why did this become a macro rather than a function?

dashboard-widgets.el
```diff
@@ -225,128 +225,100 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
+			 (dashboard-insert-shortcut ,shortcut ,section-name)
+			 max-line-length)))
 
+;;
+;; Recentf
+;;
 (defun dashboard-insert-recents (list-size)
   "Add the list of LIST-SIZE items from recently edited files."
   (recentf-mode)
-  (when (dashboard-insert-recentf-list
```
rakanalh: Dashboard let's you configure widgets to include... this is why the `when` statement is there. Why was it removed?

dashboard-widgets.el
```diff
@@ -225,128 +225,100 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
 	 "Bookmarks:"
 	 (dashboard-subseq (bookmark-all-names)
-			   0 list-size))
-    (dashboard-insert-shortcut "m" "Bookmarks:")))
+																 0 list-size)
+	 list-size
+	 "m"
+	 `(lambda (&rest ignore) (bookmark-jump ,el))
+	 (let ((file (bookmark-get-filename el)))
+								 (if file
```
rakanalh: Indentation is wrong here and the lines below

dashboard-widgets.el
```diff
@@ -225,128 +225,100 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
+						 ,list))
+		 max-line-length))
+
+(defmacro dashboard-insert-section (section-name list list-size shortcut action &rest widget-params)
+  "Add a section with SECTION-NAME and LIST of LIST-SIZE items to the dashboard.
+ACTION is theaction taken when the user activates the widget button.
+SHORTCUT is the keyboard shortcut used to access the section.
+WIDGET-PARAMS are passed to the \"widget-create\" function.
+Show EMPTY-LIST-TEXT if no items in list"
+	`(progn
```
rakanalh: Is there a specific reason of using the progn here?

dashboard-widgets.el
```diff
@@ -225,128 +225,100 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
+		 (dashboard-insert-heading ,section-name)
+		 (when-let ((max-line-length
+								 (dashboard-insert-section-list
```
rakanalh: Indentation is wrong here and the lines below

dashboard-widgets.el
jkdufair: I wanted to have each section insert return its longest line so I could do centering calculations. Each of the section and section list inserts was similar and I didn't want to just boilerplate it in there. So I refactored the section and section list inserts into a macro and replaced the repetitive sections with calls to the macro. This should also make it easy to create new sections in the future with less boilerplate code.

dashboard-widgets.el
jkdufair: I believe this still works as intended? When I remove, i.e. `'recents` from `'dashboard-items`, the section no longer appears on the dashboard. Am I understanding correctly?

dashboard-widgets.el
jkdufair: I am not aware of a way to return multiple forms to be executed in a `defmacro`. Can you advise a better method? I found nothing after a substantial amount of reading the manual, googling, etc.

dashboard.el
jkdufair: Fixed!

jkdufair: Fixed!

dashboard.el
jkdufair: Fixed!

dashboard-widgets.el
jkdufair: Fixed!

dashboard-widgets.el
jkdufair: Fixed!

dashboard-widgets.el
jkdufair: Fixed!

dashboard-widgets.el
jkdufair: Sorry. I should add there is a `when-let` in `dashboard-insert-section` that I believe accomplishes the same.

dashboard-widgets.el
jkdufair: Also it should make the intent of creating sections more clean and more declarative

dashboard-widgets.el
jkdufair: Also, the linter seems to have had trouble with `when-let` so I just did a `let` and `if`. This still appears to allow the configurability of which widgets to include.

JesusMtnez:
@jkdufair sorry for the conflict, just merge a PR that deletes `:follow-link "\C-m" which makes your macro change a conflict. :sweat_smile:

dashboard.el
```diff
@@ -62,6 +62,11 @@
   "Settings that are used in the Dashboard"
   :group 'dashboard)
 
+(defcustom dashboard-center-content t
```
rakanalh: I would rather have the default value as nil (non-centered)... might need to update the documentation with this change.

jkdufair:
I agree. I don’t want to break backward compatibility @rakanalh. The more I think about it, the less I like putting the responsibility of returning `max-line-length` on each section. If I remove that requirement, I believe custom widgets will work just fine as-is. And people can use the new macros as well to easily develop more custom widgets. Should be a simple change and it will also make the macros simpler.

Sound ok?

jkdufair:
@jesusmtnez no problem. I’ll get it fixed shortly.

JesusMtnez:
@rakanalh I don't have any possible idea to that. 

@jkdufair I like you approach, but if it is possible to keep backwards compatibility, let's try it.

In case we do not find a way, I would suggest making an stable release before merging this, so next release is a major release.

Also, good documentation is important if this change is going to include breaking changes.

jkdufair:
@rakanalh @JesusMtnez I have done the following
- Removed the need for sections to return their maximum width. I calculate it, instead, as the sections are inserted
- Verified it is backward compatible by adding a custom section
- Set the default centering to `nil` and updated the documentation accordingly
- Merged upstream master and resolved conflicts.

Thank you for your assistance with this.

jkdufair:
Glad to help. Count me in as a regular contributor.

PR #101, opened 2019-02-26
Handle the case where dashboard lines longer than win width

jkdufair:
Centering algorithm wasn't considering the fact that the longest line in the dashboard could be longer than the window width. Which isn't normally a problem. But it was trying to insert a negative amount of spaces LOL.

This fixes https://github.com/emacs-dashboard/emacs-dashboard/issues/98


JesusMtnez:
@jkdufair Thank you! I just realized that was happening this morning. :+1:

PR #102, opened 2019-02-27
Fix shortcut regression bug

jkdufair:
This also documents each shortcut so it doesn't show `??` in the keymap docs


jkdufair:
This addresses https://github.com/emacs-dashboard/emacs-dashboard/issues/97

JesusMtnez:
Thank you @jkdufair

PR #103, opened 2019-02-28
Add shortcut indicators to sections

jkdufair:
A small enhancement. Show shortcut indicators (if specified) for each section. Open to suggestions. Useful? Add font-lock?

image

### DIFF
--- a/README.org
+++ b/README.org
@@ -66,6 +66,9 @@ To update the banner or banner title
 
 ;; Content is not centered by default. To center, set
 (setq dashboard-center-content t)
+
+;; To show shortcut "jump" indicators for each section, set
+(setq dashboard-show-shortcuts t)
 #+END_SRC
 
 To customize which widgets are displayed, you can use the following snippet
--- a/README.org
+++ b/README.org
@@ -67,8 +67,8 @@ To update the banner or banner title
 ;; Content is not centered by default. To center, set
 (setq dashboard-center-content t)
 
-;; To show shortcut "jump" indicators for each section, set
-(setq dashboard-show-shortcuts t)
+;; To disable shortcut "jump" indicators for each section, set
+(setq dashboard-show-shortcuts nil)
 #+END_SRC
 
 To customize which widgets are displayed, you can use the following snippet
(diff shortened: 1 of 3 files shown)


JesusMtnez:
@jkdufair again, thank you for your contribution :smile: I really like the idea.

jkdufair:
Thanks @JesusMtnez. I'll get the linter errors fixed. Any idea if there is a better way than `(eval-when-compile (defvar...`?

jkdufair:
LOL I just moved the defcustom to the widgets file itself.

JesusMtnez:
@jkdufair what would happen if there is no keybinding for that widget (for example, custom widgets) ?

JesusMtnez:
Also @rakanalh and @jdkufair, I think it would be great if this feature is enabled by default, it does not break anything and it is a good looking thing for the package. What do you think?

jkdufair:
Ok, I will do it. Along those lines, what if we were to enable a header face by default? I just use `font-lock-function-name-face` for my own dashboard.

image

jkdufair:
@JesusMtnez I'll check on what happens with no keybinding as well. I don't think it currently handles it.

jkdufair:
Verified that custom sections without shortcuts do not cause problems.

JesusMtnez:
@jkdufair thank you so much!

JesusMtnez:
> Ok, I will do it. Along those lines, what if we were to enable a header face by default? I just use `font-lock-function-name-face` for my own dashboard.
> 
> image

We could address that in another PR :wink:

README.org
```diff
@@ -66,6 +66,9 @@ To update the banner or banner title
 
 ;; Content is not centered by default. To center, set
 (setq dashboard-center-content t)
+
+;; To show shortcut "jump" indicators for each section, set
```
JesusMtnez:
```suggestion
;; To disable shortcut "jump" indicators for each section, set
```

README.org
```diff
@@ -66,6 +66,9 @@ To update the banner or banner title
...
+(setq dashboard-show-shortcuts t)
```
JesusMtnez:
```suggestion
(setq dashboard-show-shortcuts nil)
```

PR #109, opened 2019-03-01
Feature/header faces

jkdufair:
Addresses https://github.com/emacs-dashboard/emacs-dashboard/issues/108

### DIFF
--- a/README.org
+++ b/README.org
@@ -105,11 +105,11 @@ Note that setting list-size for the agenda list is intentionally ignored; all ag
 
 It is possible to customize Dashboard's appearance using the following faces:
 
-- ~dashboard-banner-logo-title-face~ ::
+- ~dashboard-banner-logo-title~ ::
      Highlights the banner title.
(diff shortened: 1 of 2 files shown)


JesusMtnez:
@jkdufair to avoid breaking changes, take a look at: `define-obsolete-face-alias`

That way, you should be able to define an alias to the old name to indicate it is obsolete. You could set this change is included since version `1.2.6` (which will be the release containing this change).

What do you think?

jkdufair:
Oh interesting. Or do you think we are better to just leave it alone?

JesusMtnez:
I like the idea of doing things right, and we have a proper way of doing it

jkdufair:
Great. I'll get it implemented!

jkdufair:
@JesusMtnez Everything look ok?

JesusMtnez:
@jkdufair Everything looks good! Thank you!

PR #117, opened 2019-03-17
Make line moving commands behave like `dired`

rudi:
- Add `dashboard-previous-line`, `dashboard-next-line`.

- When moving point up or down a line, move to the beginning of that line’s item.  This lets us move to some line with `C-n`, then press `RET` to open that line’s item immediately.


JesusMtnez:
Thank you @rudi for this improvement! Sorry I reviewed it so late, but I've just tried it and it works perfectly!

PR #118, opened 2019-03-17
Fix error in `dashboard-next-section`

rudi:
The function body for `dashboard-next-section` is contained in the body of the starting `(interactive)` form, whereas it shouldn't be :)

The command still "works" since we try to evaluate the body of `interactive`, but pressing `}` signals an error every time.


JesusMtnez:
Thank you @rudi !!

PR #122, opened 2019-05-15
Make dashboard items list into a customizable variable

gexplorer:
I changed the **dashboard-items** variable into a customizable variable to allow dashboard items customization from within customize.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -85,13 +85,22 @@ If the value is nil then no banner is displayed.")
                                      (agenda    . dashboard-insert-agenda)
                                      (registers . dashboard-insert-registers)))
 
-(defvar dashboard-items '((recents   . 5)
-                          (bookmarks . 5)
-                          (agenda    . 5))
-  "Association list of items to show in the startup buffer.
-Will be of the form `(list-type . list-size)`.
-If nil it is disabled.  Possible values for list-type are:
-`recents' `bookmarks' `projects' `agenda' `registers'")
+(defcustom dashboard-items '((recents   . 5)
+                             (bookmarks . 5)
+                             (agenda    . 5))
+  "List of sections and items to show in the startup buffer."
+  :type '(repeat
+          (cons
+           (choice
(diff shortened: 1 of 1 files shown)


dashboard-widgets.el
```diff
@@ -85,13 +85,22 @@ If the value is nil then no banner is displayed.")
...
-                          (agenda    . 5))
-  "Association list of items to show in the startup buffer.
-Will be of the form `(list-type . list-size)`.
-If nil it is disabled.  Possible values for list-type are:
-`recents' `bookmarks' `projects' `agenda' `registers'")
+(defcustom dashboard-items '((recents   . 5)
+                             (bookmarks . 5)
+                             (agenda    . 5))
+  "List of sections and items to show in the startup buffer."
+  :type '(repeat
```
JesusMtnez: What happens here with custom widgets? If I'm understanding this `:type` section, it defines that that the items in the list can only be one of the first default widgets types, right?

gexplorer: Yes, can you give me an example of a custom widget so I can update the type?

JesusMtnez: I have an example in my own configuration: https://gitlab.com/JesusMtnez/emacs.d/blob/master/core/20-core-ui.el#L26

But everyone should be able to define their own widgets if they want.

dashboard-widgets.el
```diff
@@ -85,13 +85,22 @@ If the value is nil then no banner is displayed.")
...
+          (cons
+           (choice
+            (const :tag "Recents" recents)
+            (const :tag "Bookmarks" bookmarks)
+            (const :tag "Agenda" agenda)
+            (const :tag "Projects" projects)
+            (const :tag "Registers" registers)
+            )
+           (integer :tag "Size")
+           ))
```
jcs090218, line 102: Try not having brackets on the line on it's own.

jcs090218 (commented):
Can you fix this for us? Thanks

PR #125, opened 2019-05-21
Added init info with packages loaded and init time

ema2159:
Added init info with packages loaded and init time with an associated variable to customize the message if desired and a defcustom to enable or disable it.


JesusMtnez:
This PR should be re-done. It should only include the changes related to adding init info.

JesusMtnez:
I think the best way to proceed here is to create a new PR from current master. Closing this one.

PR #129, opened 2019-05-22
Make the icons clickable with the file names and performance improvements

### LINKED ISSUE #128: Can not select item when using icons
Using `RET` to select an item from Emacs dashboard when the cursor is at the beginning of the line, just before the icon, does not access the item.

@ema2159 maybe you can provide some insights on this one. I'll try to investigate this later.

seagle0128:
1. Fix #128: Can not select item when using icons.
1. Refactor: remove the advice for dashboard-insert-heading.
1. Use lexical binding to improve performance. 
1. Eliminate compiler warnings.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -197,15 +197,15 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
     (insert
      (cond
       ((string-equal heading "Recent Files:")
-       (all-the-icons-octicon "file-text" :height 1.2 :v-adjust 0.0 :face 'dashboard-heading))
+       (all-the-icons-octicon "file-text" :height 1.2 :v-adjust 0 :face 'dashboard-heading))
       ((string-equal heading "Bookmarks:")
-       (all-the-icons-octicon "bookmark" :height 1.2 :v-adjust 0.0 :face 'dashboard-heading))
+       (all-the-icons-octicon "bookmark" :height 1.2 :v-adjust 0 :face 'dashboard-heading))
       ((string-equal heading "Agenda for today:")
-       (all-the-icons-octicon "calendar" :height 1.2 :v-adjust 0.0 :face 'dashboard-heading))
+       (all-the-icons-octicon "calendar" :height 1.2 :v-adjust 0 :face 'dashboard-heading))
       ((string-equal heading "Registers:")
-       (all-the-icons-octicon "database" :height 1.2 :v-adjust 0.0 :face 'dashboard-heading))
+       (all-the-icons-octicon "database" :height 1.2 :v-adjust 0 :face 'dashboard-heading))
       ((string-equal heading "Projects:")
-       (all-the-icons-octicon "file-directory" :height 1.2 :v-adjust 0.0 :face 'dashboard-heading))))
+       (all-the-icons-octicon "file-directory" :height 1.2 :v-adjust 0 :face 'dashboard-heading))))
     (insert " "))
 
   (insert (propertize heading 'face 'dashboard-heading))
@@ -305,7 +305,7 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
           (dashboard-insert-ascii-banner-centered banner))))))
 
 
-(defmacro dashboard-insert-section (section-name list _list-size shortcut action &rest widget-params)
+(defmacro dashboard-insert-section (section-name list list-size shortcut action &rest widget-params)
   "Add a section with SECTION-NAME and LIST of LIST-SIZE items to the dashboard.
 SHORTCUT is the keyboard shortcut used to access the section.
 ACTION is theaction taken when the user activates the widget button.
@@ -316,7 +316,7 @@ WIDGET-PARAMS are passed to the \"widget-create\" function."
      (if ,list
          (when (dashboard-insert-section-list
                 ,section-name
-                (dashboard-subseq ,list 0 list-size)
+                (dashboard-subseq ,list 0 ,list-size)
                 ,action
                 ,@widget-params)
            (dashboard-insert-shortcut ,shortcut ,section-name))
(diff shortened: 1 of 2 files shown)


seagle0128:
My bad! I didn't realize it supports 25. Commit again.

dashboard-widgets.el
```diff
@@ -302,84 +322,54 @@ WIDGET-PARAMS are passed to the \"widget-create\" function."
...
+     (mapc
+      (lambda (el)
+        (let ((tag ,@rest))
+          (insert "\n    ")
+
+          (when (and (display-graphic-p)
+                     dashboard-set-file-icons
+                     (featurep 'all-the-icons))
+            (let* ((path (car (last (split-string ,@rest " - "))))
+                   (icon (if (file-directory-p path)
```
ema2159: In line 340 `(if (and
                                  (eq (file-remote-p path) nil)
                                  (file-directory-p path))`
should be included instead of just (if (file-directory-p path) because for some reason the file-directory-p function is what establishes unnecesary tramp connections and sometimes deadlocks or crashes the dashboard. I realized this beacause  work with remote servers a lot and I discovered this was the solution.

seagle0128: Good catch! I am not using remote server to test so didn't encounter this issue.

JesusMtnez:
Thank you @seagle0128 and @ema2159 for all your work! :+1:

dashboard-widgets.el
ema2159: Mr. seagle, I think it has to be `(eq (file-remote-p path) nil)` and not `(not (file-remote-p path)` because file-remote-p is not a boolean returning function! It returns a string if the remote server is available.

ema2159:
> Thank you @seagle0128 and @ema2159 for all your work! 👍

This is a pretty nice and useful package, its fun to work with it!

dashboard-widgets.el
ema2159: Also in line 359 the `(file-remote-p path)` doesn't return true, it returns a string so the condition is never met

seagle0128: @ema2159 I think  the check `(file-remote-p path)` is fine. The suffix `-p` implies return TRUE or FALSE. In docstring, it says `Test whether FILE specifies a location on a remote system`. Anyway, all strings including empty ("") is non-nil. So I don't see any risk here. Thoughts?

ema2159: It seems you're right.  I was confused because in the last cond statement you switched the two last conditions but I tried your code and it works!

ema2159:
Mr. @JesusMtnez, everything seems to be alright! Do your thing jaja you can merge this now

dashboard-widgets.el
seagle0128: Great! Another thing is I got a better way to resolve the issue about remote file: `(file-directory-p (file-local-name path))` has a good performance. 
I didn't change since it looks good so far. You can test on your env.

ema2159: > Great! Another thing is I got a better way to resolve the issue about remote file: `(file-directory-p (file-local-name path))` has a good performance.
> I didn't change since it looks good so far. You can test on your env.

I think I already tried that and the problem is that when the remote isn't available it deadlocks (for example, I only have access to certain remotes in my office because of proxies and stuff, and when I try to get to them in my home it deadloks)

seagle0128: Really? `file-local-name` won't trigger the remote connections IMO.

ema2159: Would you show me so I can try it out?

seagle0128: Try this:

```elisp
(defmacro dashboard-insert-section-list (section-name list action &rest rest)
  "Insert into SECTION-NAME a LIST of items, expanding ACTION and passing REST to widget creation."
  `(when (car ,list)
     (mapc
      (lambda (el)
        (let ((tag ,@rest))
          (insert "\n    ")

          (when (and (display-graphic-p)
                     dashboard-set-file-icons
                     (featurep 'all-the-icons))
            (let* ((path (file-local-name (car (last (split-string ,@rest " - ")))))
                   (icon (if (file-directory-p path)
                             (cond
                              ((and (fboundp 'tramp-tramp-file-p)
                                    (tramp-tramp-file-p default-directory))
                               (all-the-icons-octicon "file-directory" :height 1.0 :v-adjust 0.01))
                              ((file-symlink-p path)
                               (all-the-icons-octicon "file-symlink-directory"
                                                      :height 1.0 :v-adjust 0.01))
                              ((all-the-icons-dir-is-submodule path)
                               (all-the-icons-octicon "file-submodule" :height 1.0 :v-adjust 0.01))
                              ((file-exists-p (format "%s/.git" path))
                               (all-the-icons-octicon "repo" :height 1.1 :v-adjust 0.01))
                              (t (let ((matcher (all-the-icons-match-to-alist
                                                 path all-the-icons-dir-icon-alist)))
                                   (apply (car matcher) (list (cadr matcher) :v-adjust 0.01)))))
                           (cond
                            ((string-equal ,section-name "Agenda for today:")
                             (all-the-icons-octicon "primitive-dot" :height 1.0 :v-adjust 0.01))
                            ((file-remote-p path)
                             (all-the-icons-octicon "radio-tower" :height 1.1 :v-adjust 0.01))
                            (t (all-the-icons-icon-for-file (file-name-nondirectory path)))))))
              (setq tag (concat icon " " ,@rest))))

          (widget-create 'item
                         :tag tag
                         :action ,action
                         :mouse-face 'highlight
                         :button-prefix ""
                         :button-suffix ""
                         :format "%[%t%]")))
      ,list)))
```

ema2159: Ok! Let me give it a try 😁

ema2159: @seagle0128 No Mr. Zhang, for some reason this happens (this is what happened to me before also):
image
In my bookmarks there is a remote path and for some reason that implementation makes that error happen! I think it's better to leave it as it is right now. Although there are some other Centaur Emacs features that could be very nice to have like the footnote or the buttons. Although I'm not really sure what the latter would be useful for, it coul be very nice that the user was able to insert buttons with links and stuff and I think they look awesome!

seagle0128: @ema2159 I guess you are using Emacs  25.1 or 25.2. `file-local-name` is introduced in 25.2.2. It also leverage `file-remote-p` to avoid unnecessary remote connections, but makes the codes more clear.  Of course, it's fine to leave it as it is now.

```elisp
;;`file-local-name' is introduced in 25.2.2.
(unless (fboundp 'file-local-name)
  (defun file-local-name (file)
    "Return the local name component of FILE.
It returns a file name which can be used directly as argument of
`process-file', `start-file-process', or `shell-command'."
    (or (file-remote-p file 'localname) file)))
```

ema2159: No sir! I'm using emacs 26. It's really weird tho, because your implementation shouldn't give any problems.

seagle0128: You can look into the back traces and do some troubleshooting. It's easy to fix I think.

PR #131, opened 2019-05-23
Support bookmark/agenda/register/project actions.

### LINKED ISSUE #130: Bookmarks and agenda doesn't work
When a bookmark item is pressed it creates a new file instead of going to the actual bookmarked file. It seems that it's because it's taking the name of the bookmark instead of the path as the argument. When an agenda item is pressed it throws an error. This could be solved just by making agenda items non clickable. @seagle0128 mr. Zhang, please take a look at this when you're able to!

seagle0128:
Fix #130: change widget type.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -325,7 +325,7 @@ WIDGET-PARAMS are passed to the \"widget-create\" function."
 ;;
 ;; Section list
 ;;
-(defmacro dashboard-insert-section-list (section-name list _action &rest rest)
+(defmacro dashboard-insert-section-list (section-name list action &rest rest)
   "Insert into SECTION-NAME a LIST of items, expanding ACTION and passing REST to widget creation."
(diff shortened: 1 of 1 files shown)


dashboard-widgets.el
```diff
@@ -508,7 +507,7 @@ date part is considered."
    register-alist
    list-size
    "e"
-   (lambda (&rest ignore) (jump-to-register (car el)))
+   (lambda (&rest _ignore) (jump-to-register (car el)))
```
JesusMtnez: Why no `'(lambda (&rest ignore) (jump-to-register (car el)))` as in the rest of sections?

JesusMtnez:
This should also fix #133

dashboard-widgets.el
seagle0128: Yeah, others should be changed to `_ignore` as well to clean the warnings.

PR #135, opened 2019-05-24
Made heading icons a customizable variable

ema2159:
The heading icons were hard coded so I made them into a customizable alist with a a function  to modify just the icons that the user wants.


ema2159:
@JesusMtnez please check this, it's a simple modification so it shouldn't give much trouble

PR #142, opened 2019-06-03
Properly use footer icon fallback text

JesusMtnez:
Move `display-graphic-p` check to `dashboard-footer-icon` so the fallback text works properly.


JesusMtnez:
@seagle0128 this is what you meant in #141 comment, right?

seagle0128:
Right! Thank you!

PR #147, opened 2019-06-04
Support `straight.el` properly in `dashboard-init-info`

JesusMtnez:
- Use `(bound-and-true-p package-alist)` to check if `package.el` is enabled.
- Use ` (and (boundp 'straight--profile-cache) (hash-table-p straight--profile-cache))` to check if `straight.el` is enabled.

This should fix discussion in https://github.com/emacs-dashboard/emacs-dashboard/pull/138

@romatthe could you please check if this works for you with `straight.el`?
@ema2159 I would also like your opinion on this

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -93,7 +93,10 @@ to the specified width, with aspect ratio preserved."
   (if (bound-and-true-p package-alist)
       (format "%d packages loaded in %s"
               (length package-activated-list) (emacs-init-time))
-    (format "Emacs started in %s" (emacs-init-time)))
+    (if (and (boundp 'straight--profile-cache) (hash-table-p straight--profile-cache))
+        (format "%d packages loaded in %s"
+                (hash-table-size straight--profile-cache) (emacs-init-time))
+      (format "Emacs started in %s" (emacs-init-time))))
   "Init info with packages loaded and init time.")
 
 (defvar dashboard-footer
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -90,7 +90,7 @@ to the specified width, with aspect ratio preserved."
 
 (defvar dashboard-init-info
   ;; Check if package.el was loaded and if package loading was enabled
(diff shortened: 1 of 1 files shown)


ema2159:
Let me give it a try!

raxod502:
That seems less than optimal. What about `(bound-and-true-p package-alist)`? That will trigger if and only if `package.el` is loaded _and_ initialized _and_ at least one package was activated.

Of course, this only works after startup.

JesusMtnez:
May I know why it is less than optimal? I'm not an expert on elisp

raxod502:
Sure, no problem. Your code will trigger just because `package.el` was loaded, even if it's not being used. My code avoids that problem, since `package-alist` remains at nil until at least one package is actually loaded by `package.el`.

JesusMtnez:
Thank you @raxod502 for the explanation. I'll update this PR then

raxod502:
If you want to support `straight.el`, you could test for `(and (boundp 'straight--profile-cache) (hash-table-p straight--profile-cache))` and then check `(hash-table-size straight--profile-cache)` to get the number of packages. That said, this is unsupported and there really ought to be a supported way to get information about installed packages from `straight.el`.

JesusMtnez:
Thank you @raxod502 for your suggestions, I've applied them so we could provide a proper support not only for `package.el` but also `straight.el`. Does it look good now?

@romatthe could you also check if this works well for you ?

ema2159:
@JesusMtnez  this doesn't seem to work either. What could be the problem?

raxod502:
I don't use this package, but evaluating that variable definition in my Emacs session does indeed give

```
"145 packages loaded in 0.8 seconds"
```

ema2159:
> I don't use this package, but evaluating that variable definition in my Emacs session does indeed give
> 
> ```
> "145 packages loaded in 0.8 seconds"
> ```

You're right, I forgot to byte compile jaja it does work.

PR #149, opened 2019-06-06
Add newline at the end of insert footer.

jcs090218:
I think you should have a line at the end of the `*dashboard*` buffer?


ema2159:
Hi! May I ask, why is this necessary?

jcs090218:
Before the version dashboard has footer implemented, dashboard buffer has a very bottom line at the end of the buffer. So I think it is good to keep it that way?

Issue #153, opened 2019-06-16
Allow custom ASCII startup banner too, not only PNG


andrewbanchich:
I see from the README: `;; 1, 2 or 3 which displays one of the text banners` but how can I add my own?

JesusMtnez:
As far as I know, you would need to added inside the package `banner` folder: https://github.com/emacs-dashboard/emacs-dashboard/tree/master/banners

Maybe a new feature request could be passing the path to a `png` file or `txt` file and handle those cases, since right now it only supports `png` files when passing a path as parameter for `dashboard-startup-banner`.

andrewbanchich:
Should I open a new feature request issue for this then?

jcs090218:
Would it be easier to just have a variable that point to the path of the text file?

JesusMtnez:
> Should I open a new feature request issue for this then?

Let's keep these issue to track that feature. I'll rename it.

> Would it be easier to just have a variable that point to the path of the text file?

We could use the same variable to point to the banner we want to use, and make a logic so it could use both types correctly. It shouldn't be difficult. I'll try to work on that when I get some time :sweat:

krzkrzkrz:
Would like this too. Any updates here?

Hkyx:
you can create a 4.txt and put your ascii in my case it work.

jcs090218:
> you can create a 4.txt and put your ascii in my case it work.

I think replacing the current path is a work around. But it would be great we can point to the file directly! 😄

Hkyx:
yeah but I think it's probably not the most important think, like I got a good config file and if I save it and refresh dashboarfd all works, If I close/open emacs, dashboard use only the half of the config file. so a txt file to create is a little think ;p

GiantToast:
Hey guys, I wanted this too and stumbled upon this issue.
I ended up solving it by creating a banners folder in my `.emacs.d` folder, in which I created a custom `4.txt` banner.
Then I modified my dashboard elisp to be the following:
```elisp
(setq user-banners-dir (concat user-emacs-directory (convert-standard-filename "banners/")))

(defun install-banners ()
  "Copy all files under under banners directory to dashboard banners directory"
  (when (boundp 'dashboard-banners-directory)
    (copy-directory user-banners-dir dashboard-banners-directory nil nil t)))

(use-package page-break-lines
  :ensure t)

(use-package dashboard
  :ensure t
  :after (page-break-lines)
  :config
  (install-banners)
[...]

PR #157, opened 2019-06-21
Text startup banner.

jcs090218:
This is a quick implementation for text banner. See #153. The benefits are this is down compatible if the user want to use default `1.txt`, `2.txt` and `3.txt` banner files. Otherwise, they can just set the `dashboard-banner-logo-text` variable like the line of code below. If they want the default banner just set to nil.

```el
(setq dashboard-banner-logo-text "dir/banner.txt")
```

**Edit:** This implementation may not be the best solution, because I don't know what the author of `dashboard` wants. So I guess this PR is just an advice.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -86,9 +86,11 @@ to the specified width, with aspect ratio preserved."
   (expand-file-name (concat dashboard-banners-directory "logo.png"))
   "Emacs banner image.")
 
-(defvar dashboard-banner-logo-text
+(defcustom dashboard-banner-logo-text
   (expand-file-name (concat dashboard-banners-directory "1.txt"))
-  "Emacs banner text.")
+  "Emacs banner text."
+  :type 'string
+  :group 'dashboard)
 
 (defconst dashboard-banner-length 75
   "Width of a banner.")
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -86,6 +86,10 @@ to the specified width, with aspect ratio preserved."
   (expand-file-name (concat dashboard-banners-directory "logo.png"))
   "Emacs banner image.")
 
+(defvar dashboard-banner-logo-text
+  (expand-file-name (concat dashboard-banners-directory "1.txt"))
+  "Emacs banner text.")
+
 (defconst dashboard-banner-length 75
   "Width of a banner.")
 
@@ -353,7 +357,10 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
 
 (defun dashboard-get-banner-path (index)
   "Return the full path to banner with index INDEX."
-  (concat dashboard-banners-directory (format "%d.txt" index)))
+  (if (and dashboard-banner-logo-text
+           (file-exists-p dashboard-banner-logo-text))
+      dashboard-banner-logo-text
+    (concat dashboard-banners-directory (format "%d.txt" index))))
 
 (defun dashboard-choose-banner ()
   "Return the full path of a banner based on the dotfile value."
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -141,7 +141,7 @@ Example:
 Default value is `official', it displays
 the Emacs logo.  `logo' displays Emacs alternative logo.
 An integer value is the index of text
-banner.  A string value must be a path to a .PNG file.
+banner.  A string value must be a path to a .PNG or .TXT file.
 If the value is nil then no banner is displayed.")
 
 (defvar dashboard-buffer-last-width nil
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -86,12 +86,6 @@ to the specified width, with aspect ratio preserved."
   (expand-file-name (concat dashboard-banners-directory "logo.png"))
   "Emacs banner image.")
 
-(defcustom dashboard-banner-logo-text
-  (expand-file-name (concat dashboard-banners-directory "1.txt"))
-  "Emacs banner text."
-  :type 'string
-  :group 'dashboard)
-
 (defconst dashboard-banner-length 75
   "Width of a banner.")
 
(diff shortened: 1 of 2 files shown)


dashboard-widgets.el
```diff
@@ -353,7 +357,10 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
 
 (defun dashboard-get-banner-path (index)
   "Return the full path to banner with index INDEX."
-  (concat dashboard-banners-directory (format "%d.txt" index)))
+  (if (and dashboard-banner-logo-text
+           (file-exists-p dashboard-banner-logo-text))
+      dashboard-banner-logo-text
+    (concat dashboard-banners-directory (format "%d.txt" index))))
```
jcs090218: Check if there are value in `dashboard-banner-logo-text` and ensure the file exists, otherwise use the default banner.

dashboard-widgets.el
```diff
@@ -86,6 +86,10 @@ to the specified width, with aspect ratio preserved."
   (expand-file-name (concat dashboard-banners-directory "logo.png"))
   "Emacs banner image.")
 
+(defvar dashboard-banner-logo-text
```
jcs090218: I don't know what I should use here because I saw `defcustom` and `defvar` everywhere between the user customize code and package code. So here I simply used `defvar`.

JesusMtnez:
@jcs090218 thank you so much for this work! Every pice of work is always welcome :smile: 

Have you tried it? I've tried to use this branch but it didn't work for me. Even if I set up `dashboard-logo-banner-text` to an ascii version of Emacs logo, it didn't use it at all.

The problem is in the `dashboard-choose-banner`. Even if you setup correctly `dashboard-logo-banner-text`, since its value is not an integer, it would never meet the condition to choose to the text banner.

I think the key of this feature is not about adding a new variable, but modifying `dashboard-choose-banner` to check if the file extension is `txt` and loading it as a text banner.

Feel free to ask any doubts you may have.

dashboard-widgets.el
JesusMtnez: Since this value should be customizable, I think it should be a `defcustom`.

JesusMtnez:
Also, remember that adding new features / improving current ones may require an update of the documentation too :wink:

jcs090218:
> Have you tried it? I've tried to use this branch but it didn't work for me. Even if I set up dashboard-logo-banner-text to an ascii version of Emacs logo, it didn't use it at all.

I did test it, and it works for me. Did you try it in the terminal? Cuz this would only works inside terminal? :/

> The problem is in the dashboard-choose-banner. Even if you setup correctly dashboard-logo-banner-text, since its value is not an integer, it would never meet the condition to choose to the text banner.

Not quite sure what do you mean by this. In the function `dashboard-choose-banner` if the value's isn't integer wouldn't matter, wouldn't it? Cuz the code eventually goes down to `dashboard-get-banner-path` by default.

> I think the key of this feature is not about adding a new variable, but modifying dashboard-choose-banner to check if the file extension is txt and loading it as a text banner.

Thanks for letting me know! I will do this when I have time! :D

> Also, remember that adding new features / improving current ones may require an update of the documentation too 😉

And this too!

dashboard-widgets.el
jcs090218: Resolved!

dashboard-widgets.el
jcs090218: Removed!

dashboard-widgets.el
```diff
@@ -368,10 +368,7 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
              (dashboard-get-banner-path 1)))
           ((integerp dashboard-startup-banner)
            (dashboard-get-banner-path dashboard-startup-banner))
-          ((and dashboard-startup-banner
-                (image-type-available-p (intern (file-name-extension
-                                                 dashboard-startup-banner)))
-                (display-graphic-p))
+          ((stringp dashboard-startup-banner)
```
jcs090218: This might sounds stupid, but I changed to check if the value is string and ignore to check if it is image and `display-graphic-p`. So just let the user decide what file to load? Just mentioned prefer setting it this way.

```el
(setq dashboard-startup-banner (if (display-graphic-p) "path/to/image.png" "path/to/text.txt")
```

dashboard-widgets.el
```diff
@@ -141,7 +141,7 @@ Example:
 Default value is `official', it displays
 the Emacs logo.  `logo' displays Emacs alternative logo.
 An integer value is the index of text
-banner.  A string value must be a path to a .PNG file.
+banner.  A string value must be a path to a .PNG or .TXT file.
```
jcs090218: Just strictly ask the user to have `.TXT` or `.PNG` file.

jcs090218:
@JesusMtnez What do you think about this patch?

JesusMtnez:
@jcs090218 Sorry, but I couldn't take a look. I'll try to review it before next week to give you some feedback. Thank you for your patience.

jcs090218:
@JesusMtnez Any progress on this patch? :D

JesusMtnez:
@jcs090218 Sorry, I'm pretty busy these days. I'll try to review it this week.

jcs090218:
@JesusMtnez No worries, I don't mean to rush you! :D

dashboard-widgets.el
JesusMtnez: I like your suggestion on how to setup `dashboard-startup-banner`, I would like to see the `README.md` updated with that. 

But this line change would break backwards compatibility. Until now, people set `dashboard-startup-banner` like this:
```elisp
(dashboard-startup-banner "/tmp/emacs-dashboard/banners/logo.png")
```

If you start emacs in terminal with the variable like set like that, it would fail to start. I think you should handle what happens if the value is a `png` file but emacs is started with now window system.

image

JesusMtnez:
@jcs090218 Any plans on updating this work to latest `emacs-dashboard`? :thinking:

jcs090218:
I think the latest post is about the review? The conflict must have a write access to this branch hence I couldn't resolve it! You might have to resolve or else I would have open another PR? Thanks! 😄

jcs090218:
Move to #246. Close this now!

PR #165, opened 2019-06-28
Remove `(require 'seq)`

jcs090218:
Is this something what you want? #161

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -20,7 +20,6 @@
 
 ;;; Code:
 
-(require 'seq)
 (require 'page-break-lines)
 (require 'recentf)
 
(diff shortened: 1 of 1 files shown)


jcs090218:
@JesusMtnez Does this works for you? Cuz after I updated to this branch, it breaks again... Sorry for this PR. Not sure where are the differences between `seq-take` and the code implemented...

dashboard.el
```diff
@@ -145,16 +144,21 @@ Optional prefix ARG says how many lines to move; default is one line."
...
+        (max-line-length 0)
+        (recentf-items '()))
     ;; disable recentf mode,
     ;; so we don't flood the recent files list with org mode files
     ;; do this by making a copy of the part of the list we'll use
     ;; let dashboard widgets change that
     ;; then restore the orginal list afterwards
     ;; (this avoids many saves/loads that would result from
     ;; disabling/enabling recentf-mode)
-    (if recentf-is-on
```
jcs090218: If you are going to restore this, you might want to use `when` here.

JesusMtnez:
@jcs090218 how does it break? I'm not having any problem :thinking:

jcs090218:
@JesusMtnez Sorry for reporting this. I guess is just a versioning issue on my company's PC. :P My bad!

JesusMtnez:
I had to revert the change, something was wrong when the emacs-dashboard was refreshed. :thinking:

jcs090218:
I was confused, I guess is because I overwrite the dashbaord refresh function. Sorry about this PR. :(

JesusMtnez:
No worries @jcs090218 ! These things happen sometimes, even more often when there is not a test suite to check those changes. But I think this will help us with the next PR to completely remove `(require 'seq)` for sure :+1:

PR #167, opened 2019-06-29
Added filtering by category for org mode

SidharthArya:
Added support for filtering by category in dashboard, so the users can now select the org categories which they wants to have in the dashboard. 
The users would have to specify the category which they need in their dashboard, or if they don't all categories will be considered.

So in my config I have something like this:

```
(use-package dashboard
  :ensure t
  :config
(setq dashboard-items '((recents . 10) (bookmarks . 5) (agenda . 10)))
  (setq dashboard-org-agenda-categories '("Tasks"))
 
```

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -73,6 +73,13 @@ to the specified width, with aspect ratio preserved."
   :type 'boolean
   :group 'dashboard)
 
+(defcustom dashboard-org-agenda-categories nil
+  "Specify the Categories to consider when using agenda in dashboard.
+Example:
+'(\"Tasks\" \"Habits\")"
+  :type 'list
+  :group 'dashboard)
+
 (defconst dashboard-banners-directory
   (concat (file-name-directory
            (locate-library "dashboard"))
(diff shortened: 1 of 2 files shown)


dashboard-widgets.el
```diff
@@ -99,6 +99,13 @@ The format is: 'icon title help action face prefix suffix'.
 Example:
 '((\"☆\" \"Star\" \"Show stars\" (lambda (&rest _) (show-stars)) 'warning \"[\" \"]\"))")
 
+(defvar dashboard-org-agenda-categories nil
```
JesusMtnez: Since `dashboard-org-agenda-categories` its something that the user may need to customize, I think it should be a `defcustom`.

SidharthArya: Yes, thank you. I have made the changes you suggested.

dashboard-widgets.el
```diff
@@ -99,6 +106,8 @@ The format is: 'icon title help action face prefix suffix'.
 Example:
 '((\"☆\" \"Star\" \"Show stars\" (lambda (&rest _) (show-stars)) 'warning \"[\" \"]\"))")
 
+
```
JesusMtnez: Any reason to add those two empty lines (109 and 110) ?

JesusMtnez:
Thank you @SidharthArya !

Issue #146, opened 2019-06-03
7 day agenda view


miwinning:
Using (setq show-week-agenda-p t) for the 7 day agenda view does not work with (setq dashboard-set-heading-icons t).

It causes an error:- "Wrong type argument: char-or-string-p, nil"

This happens when run in GUI but not with emacs -nw

ema2159:
I think we had this problem previosly or a similar bug, I'll try to see what's going on

alecvn:
Can confirm I have the same problem.  Is there any solution for this yet?

JesusMtnez:
Could you confirm #173 fixed the problem? I'm not an org-agenda user :sweat:

miwinning:
Confirmed fix works

JesusMtnez:
Thank you @DuCorey for the fix
Thank you @miwinning for the confirmation!

Great job!

PR #173, opened 2019-07-02
Fix icons for weekly agenda and add an icon final condition

DuCorey:
This PR is a fix for #146. 

I simply check for either the daily agenda or the weekly agenda string. 

I've also taken the liberty of adding a final condition that will always return an empty string so that the heading creation cannot fail when it can't find an icon. This could be useful for people who want to make a custom heading and still some of the icons.


DuCorey:
My bad for the linting. I forgot to run the tests on my machine first.

JesusMtnez:
Thank you @DuCorey !

PR #175, opened 2019-07-10
Remove page-break-lines support.

Luis-Henriquez-Perez:
There were only 2 lines where page-break-lines was used. The only potential issue with this PR that I see is that it might be slightly less convenient for some users who use the `page-break-lines` support.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -151,7 +151,7 @@ Example:
                 "Welcome to the church of Emacs"
                 "While any text editor can save your files,\
  only Emacs can save your soul"
-                "I showed you my source code,pls respond"
+                "I showed you my source code, pls respond"
                 )))
     (nth (random (1- (1+ (length list)))) list))
   "A footer with some short message.")


JesusMtnez:
@Luis-Henriquez-Perez I like the changes introduced by your last commit, but I would like to keep them in another PR, please :smile:

Luis-Henriquez-Perez:
Ok, sorry about that. I haven't been PRing for long so I'm not sure of how to create a separate pull request with a different commit. I'll look into that. Ideally I wanted each one of these commits to be separate PRs.

Luis-Henriquez-Perez:
I've reverted it to the commit relevant to this issue. I will make separate issues for the other commits.

zjp:
The code that inserts page breaks is still there, so removing the page-break-lines-mode dependency just makes the dashboard display them as ^L -- imo it might be better to make it optional and then hide the code that prints the page breaks behind a check for whether the user has page-break-lines enabled in their configuration.

Luis-Henriquez-Perez:
> The code that inserts page breaks is still there

This was intentional. I don't want to remove the feature. I just want it so that
dashboard does not break if `page-break-lines` is not installed.

> just makes the dashboard display them as ^L

I was thinking dashboard could display a set of dashes the length of the page
width if the user chooses not to use `page-break-lines`. Since, after all, the
`^L` is just a remant of `page-break-lines`. And it looks bad.

> it might be better to make it optional and then hide the code that prints
the page breaks behind a check for whether the user has page-break-lines enabled in their configuration

Yes, I also think we should make it optional. Perhaps an option like
`dashboard-enable-page-break-lines`. I don't agree with having a check for
whether the user has `page-break-lines` enabled because I don't think we should
assume that if the user has that package enabled, they want to use it for the
dashboard.

zjp:
> > The code that inserts page breaks is still there
> 
> This was intentional. I don't want to remove the feature. I just want it so that
> dashboard does not break if `page-break-lines` is not installed.
I see. I agree with that. 
> > just makes the dashboard display them as ^L
> 
> I was thinking dashboard could display a set of dashes the length of the page
> width if the user chooses not to use `page-break-lines`. Since, after all, the
> `^L` is just a remant of `page-break-lines`. And it looks bad.
That's a good idea, too. 
> > it might be better to make it optional and then hide the code that prints
> > the page breaks behind a check for whether the user has page-break-lines enabled in their configuration
> 
> Yes, I also think we should make it optional. Perhaps an option like
> `dashboard-enable-page-break-lines`. I don't agree with having a check for
> whether the user has `page-break-lines` enabled because I don't think we should
> assume that if the user has that package enabled, they want to use it for the
> dashboard.
That's what I meant by 'check whether the user has it enabled'. The function that prints the page break is in `dashboard-widgets.el` -- perhaps print the dashes if `dashboard-enable-page-break-lines` is unset, and page breaks if some form of `(setq dashboard-enable-page-break-lines t)` is in the user's config?

Luis-Henriquez-Perez:
> perhaps print the dashes if dashboard-enable-page-break-lines is unset, and page breaks if some form of (setq dashboard-enable-page-break-lines t) is in the user's config?

That sounds good. Only thing I would add I think there should also be a way for the user to have nothing--no dashes or page-break-lines. Perhaps this should be the default?

I think probably having nothing as a default is the best option, while allowing the user to optionally add dashes, or page break lines. However, is it OK to change the default like this? I say this because right now the "default" is `page-break-lines`. How much do we care about potentially causing an error to a users dashboard when they update it?

zjp:
> > perhaps print the dashes if dashboard-enable-page-break-lines is unset, and page breaks if some form of (setq dashboard-enable-page-break-lines t) is in the user's config?
> 
> That sounds good. Only thing I would add I think there should also be a way for the user to have nothing--no dashes or page-break-lines. Perhaps this should be the default?

I like this, too. 

> I think probably having nothing as a default is the best option, while allowing the user to optionally add dashes, or page break lines. However, is it OK to change the default like this? I say this because right now the "default" is `page-break-lines`. How much do we care about potentially causing an error to a users dashboard when they update it?

I can't answer that unfortunately; it depends on what the maintainer ends up saying.

Issue #176, opened 2019-07-14
Some places not using dashboard-buffer-name


vermiculus:
https://github.com/emacs-dashboard/emacs-dashboard/blob/f76be664ef3e890de03816fd8a2f7e3f74f04982/dashboard.el#L250

https://github.com/emacs-dashboard/emacs-dashboard/blob/f76be664ef3e890de03816fd8a2f7e3f74f04982/dashboard-widgets.el#L255

this is causing problems when customizing the dashboard buffer name -- specifically, a buffer `*dashboard*` still gets created, but is mostly empty (save a few `^L`s).

jcs090218:
This is completed by eca733a829c5b37ae770a0c881b4c50f7dd646c0 and 02cf7b871c822461ca9063505a3de6a0f79499af.

PR #178, opened 2019-07-20
Use dashboard-buffer-name everywhere when invoking buffers

sehnsucht13:
This takes care of #176 and makes it easier to customize for users.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -252,7 +252,7 @@ Optionally, provide NO-NEXT-LINE to move the cursor forward a line."
 (defun dashboard-append (msg &optional _messagebuf)
   "Append MSG to dashboard buffer.
 If MESSAGEBUF is not nil then MSG is also written in message buffer."
-  (with-current-buffer (get-buffer-create "*dashboard*")
+  (with-current-buffer (get-buffer-create dashboard-buffer-name)
     (goto-char (point-max))
     (let ((buffer-read-only nil))
       (insert msg))))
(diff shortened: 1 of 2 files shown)


sehnsucht13:
The lint check does not seem to like the free use of the variable. Looking at the code, there seem to be two options:
1. Ignore the lint check for the free variable. Since dashboard-widgets is required from dashboard.el and the variable is defined inside dashboard.el, there will be no problems loading this at all.
2. Move the offending function which is called `dashboard-append` from `dashboard-widgets.el` to `dashboard`.

Please let me know which is preferable or any other options and I will make sure to implement them. Thanks

JesusMtnez:
@sehnsucht13 Thank you for this change!

Another solution could be:
- Update this branch with latest `master` change.
- Add `(defvar dashboard-buffer-name)` [here](https://github.com/emacs-dashboard/emacs-dashboard/blob/7a71e6ca4c32fdadde0c8624ea4e2e7c11474e7d/dashboard-widgets.el#L23)

sehnsucht13:
Thanks a lot for the quick response and suggestions! I will try to take care of this tommorow!

jcs090218:
Any progress on this? 😕

jcs090218:
Ping @sehnsucht13, I think there is still an error in CI.

jcs090218:
This is completed by eca733a829c5b37ae770a0c881b4c50f7dd646c0 and 02cf7b871c822461ca9063505a3de6a0f79499af.

PR #182, opened 2019-08-18
Disable display-line-numbers-model

zjp:
This is a short ~~one~~two-liner commit to add support for hiding line numbers by default for users of `global-display-line-numbers-mode`, the way they are hidden for users of `global-linum-mode`. It checks the major version first, since the new mode was added in Emacs 26.

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -56,6 +56,8 @@
   (buffer-disable-undo)
   (whitespace-mode -1)
   (linum-mode -1)
+  (when (>= emacs-major-version 26)
+    (display-line-numbers-mode -1))
   (page-break-lines-mode 1)
   (setq inhibit-startup-screen t)
   (setq buffer-read-only t


jcs090218:
There are few problems with this PR.

1. There are something wrong with CI, maybe you should fix it.
2. Use `when` if you only have one condition.

zjp:
It looks like now there's an issue with the final step that isn't due to this PR in particular.

jcs090218:
Yeah, I think you could just update the branch to the latest commit to resolve this kind of issue. Anyway, I am good with this kind of PR. Thanks!

One last minor issue.
* On line 59, indentation should be 4 not 6.

zjp:
I have rebased on master and fixed spacing. I'm not sure why test-emacs-25 is failing; when a condition fails the following elisp isn't supposed to be evaluated, so Emacs 25 shouldn't even know that symbol is there. Is this not the case for `when`?

jcs090218:
@zjp I don't think the `when` operator shouldn't do to anything like this. The log says this

```sh
emacs --batch -q -l .emacs/init.el
Contacting host: elpa.gnu.org:80
Failed to download `melpa' archive.
Package `use-package-' is unavailable
Makefile:12: recipe for target '.elpa' failed
make: *** [.elpa] Error 255
```

I think we could simply just rerun the test. @JesusMtnez What do you think? 😕 

Thanks!

PR #192, opened 2019-09-22
fix "invalid face property: inherit nil" on navigator icons

mohkale:
this request addresses issue [191](https://github.com/emacs-dashboard/emacs-dashboard/issues/191). It's a small change and logically identical to before, only now there's an extra `nil` check to ensure the text isn't propertized with an invalid face property. 😄.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -440,10 +440,11 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
           (widget-create 'item
                          :tag (concat
                                (when icon
-                                 (propertize icon 'face `(:inherit
-                                                          ,(get-text-property 0 'face icon)
-                                                          :inherit
-                                                          ,face)))
+                                 (propertize icon 'face
+                                             (let ((prop-face (get-text-property 0 'face icon)))
+                                               (if prop-face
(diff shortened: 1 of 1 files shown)


mohkale:
I don't think this is ever gonna be merged, so I'll just delete it here.
If anyone else ever ends up facing this issue, you can get around it by
propertizing whatever string you had, that wasn't propertized, with a
non-nil face attribute.

This'll almost certainly be a plain string, which you've tried to use in
place of an icon.

eg. the following works:

```elisp
(setq dashboard-navigator-buttons
      `(((,(propertize "?" 'face '(:inherit default)) ;; needs a face attribute
          "" "emacs FAQ" #'view-emacs-faq font-lock-builtin-face))))
```

JesusMtnez:
Sorry I have not been very active lately. I think it is a valuable pr for the project, I'll review it next week so we can merge it.

mohkale:
@JesusMtnez sorry for the rushed close, I've spent the day cleaning up my account and seeing as theirs a workaround for this issue I decided to close the PR. If you feel it's worth merging that's great, but there's no huge rush; feel free to take your time 😄.

jcs090218 (commented):
I have checked, I think this can be merged!

PR #202, opened 2019-12-11
Make dashboard shortcuts configurable

winny-:
- Makes shortcuts configurable
- Only show shortcuts that are not `nil`
- Use cleaner identifiers for the shortcut lambdas
- Add docstrings to the shortcut lambda (visible with e.g. `C-h k p`)

I'm an elisp novice, any suggestions appreciated :)

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -230,6 +230,19 @@ If nil it is disabled.  Possible values for list-type are:
   :type  '(repeat (alist :key-type symbol :value-type integer))
   :group 'dashboard)
 
+(defcustom dashboard-item-shortcuts '((recents . "r")
+                                      (bookmarks . "m")
+                                      (projects . "p")
+                                      (agenda . "a")
+                                      (registers . "e"))
+  "Association list of items and their corresponding shortcuts.
+Will be of the form `(list-type . keys)' as understood by
+`(kbd keys)'.  If nil, shortcuts are disabled.  If an entry's
+value is nil, that item's shortcut is disbaled.  See
+`dashboard-items' for possible values of list-type.'"
+  :type '(repeat (alist :key-type symbol :value-type string))
+  :group 'dashboard)
+
 (defcustom dashboard-items-default-length 20
   "Length used for startup lists with otherwise unspecified bounds.
 Set to nil for unbounded."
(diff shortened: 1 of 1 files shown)


winny-:
Something I wasn't able to get working quite right is: `C-h f dashboard-jump-to-recent-files RET`. Emacs says `user-error: Symbol’s function definition is void: dashboard-jump-to-recent-files`.

I don't think it is completely necessary to be able to get this working, though it would be what the user expects. Any suggestions appreciated :)

dashboard-widgets.el
```diff
@@ -254,15 +265,30 @@ Return entire list if `END' is omitted."
...
   `(progn
      (eval-when-compile (defvar dashboard-mode-map))
-     (let ((sym (make-symbol (format "Jump to \"%s\"" ,search-label))))
+     (let ((sym (make-symbol (format "dashboard-jump-to-%s" (downcase
+                                                             (replace-regexp-in-string
+                                                              " +" "-" (replace-regexp-in-string
+                                                                        ":+" "" ,search-label
+                                                                        nil nil nil)
+                                                              nil nil nil))))))
        (fset sym (lambda ()
```
kisaragi-hiu: This function definition happens within the `let` during runtime—it is thrown away after the runtime `let` exits. This is why `C-h f dashboard-jump-to-recent-files` fails. This needs to be replaced with a `defun`, like:

```elisp
  (let ((sym (make-symbol (format "dashboard-jump-to-%s" (downcase search-label)))))
    `(progn
       (eval-when-compile (defvar dashboard-mode-map))
       (defun ,sym nil
         ,(concat
           "Jump to "
           (replace-regexp-in-string
            ":+" "" (format "%s" search-label) nil nil nil)
           ".  This code is dynamically generated in `dashboard-insert-shortcut'.")
         (interactive)
         (unless (search-forward ,search-label (point-max) t)
           (search-backward ,search-label (point-min) t))
         ,@(unless no-next-line
             '((forward-line 1)))
         (back-to-indentation))
```

winny-: Thanks for the suggestion! I tried this out in another branch, but was not able to get it working. The code works, but indeed I am not able to do either `M-x dashboard-jump-to-projects RET` or `C-h f dashboard-jump-to-project RET` (with the same error as above).

https://github.com/winny-/emacs-dashboard/commit/0e0a56da0b19fbb00a8618119d3ed9d604b0de0c

kisaragi-hiu: Looking at your linked commit, the problem is that you're using `make-symbol`, which gives you an uninterned symbol, ie. one that is unique. You need to use `intern` instead.

```elisp
(eq 'testing (intern "testing")) ; => t
(eq 'testing (make-symbol "testing")) ; => nil
```

winny-: Thank you very much for the help! That did the trick! =)

winny-:
Applied the changes suggested, let me know if there are any other things to adjust =)

winny-:
Hello. Is there interest to merge this feature in? Does this need further review?

jcs090218:
This is neat. I like the idea. I don't think merging this would have any issue.

I will make a quick test next week. Thanks for the PR!

jcs090218:
I have tested this PR and it works well!

Once thing I might worry is that the other extension that extends this package might want to know this; yet there is no conflict to this.

jcs090218 (approved):

jcs090218:
Merged! Thank you!

winny-:
Cheers.  Congrats on the project transition!

PR #211, opened 2020-02-10
Smaller icon size for remote files

seagle0128:
The icon `radio-tower` is too big.


jcs090218:
Maybe we can set this as customizable? So we don't have to tweak this manually. Just let the user tweak it for themselves. 😕

seagle0128:
Maybe, but it's good enough for now. That feature can be implemented in future.

Issue #208, opened 2020-01-12
Optional activation of dashboard


csevast:
I want to adapt my `init.el` in a way the `dashboard` to start optionally. I want to have different Emacs instances, with or without `dashboard`, using the same initialization file. Unfortunately, when the `dashboard-setup-startup-hook` is evaluated after Emacs' start-up nothing happens. Is there any `dashboard` function to start displaying it's buffer after Emacs' start-up?

jcs090218:
Try this.

```el
(defun my--startup-dashboard()
  (switch-to-buffer "*dashboard*")
  (goto-char (point-min))
  (redisplay))
```

jcs090218:
@csevast Does the code work for you? Thanks!

jcs090218:
I think this is resolved? Close this issue now! :)

PR #214, opened 2020-02-16
Optional activation of dashboard, impls to #208.

jcs090218:
Just the implementation for #208.


jcs090218:
This can be further improve. Close this for now. ;)

PR #222, opened 2020-02-26
Check if *dashboard* buffer exists or not.

nasyxx:
Sometimes we may have accidentally killed the `\*dashboard\*` buffer.  And then `(kill-buffer dashboard-buffer-name)` will not work and give out an error.  Thus, when the buffer not exists, we can ignore the `kill-buffer` step.


jcs090218:
Make sense, merged!

PR #227, opened 2020-05-16
Allow moving the point during dashboard-mode-hook

Dragoncraft89:
If the point is being moved in the dashboard-mode-hook, e.g. moved to the first entry, this has no effect as the point is moved to the beginning of the buffer after the hook.

The point is again moved back to the beginnig during the `emacs-startup-hook`. This one isn't necessary though as the point is not being moved in between.


jcs090218 (approved):

jcs090218:
> The point is again moved back to the beginnig during the emacs-startup-hook. This one isn't necessary though as the point is not being moved in between.

You mean the point is automatically moved to beginning of the buffer in `emacs-startup-hook`? I think we are just ensuring the point start at the beginning of the buffer.

Dragoncraft89:
> You mean the point is automatically moved to beginning of the buffer in emacs-startup-hook?

Exactly

JesusMtnez (approved):

JesusMtnez:
@Dragoncraft89 Thank you so much for the contribution!!

jcs090218:
I haven't test this patch yet. I will test it with in few days.

jcs090218:
Merged! Thanks!

Issue #215, opened 2020-02-16
Support for counsel-projectile


yurikoif:
Dashboard is amazing, but it supports projectile rather than its variation counsel-projectile.
When I try to open a project, the file list is shown in a raw projectile-style single line mini-buffer, but what I want is a counsel(ivy)-style mini-buffer.

jsigman:
Also came here looking for this!

mohkale:
not sure what you mean? Could you post a screenshot of what you'd like to be shown and what is actually shown?

yurikoif:
> not sure what you mean? Could you post a screenshot of what you'd like to be shown and what is actually shown?

Thank you for replying.
Selection minibuffer in cousel-projectile looks like this:
![Screenshot from 2020-06-25 05-19-44 (1)](https://user-images.githubusercontent.com/10281921/85693651-6d211d80-b6a4-11ea-9ab0-2e9496066189.png)
while dashboard uses original ido style selection minibuffer:
`[project-name] Find file: {a | b | c | ...}`

mohkale:
which command are you running to get a list of projects to switch to?

These are all the commands my dashboard install has available:

```
    dashboard-mode
    dashboard-return
    dashboard-next-line
    dashboard-a-shortcut
    dashboard-p-shortcut
    dashboard-r-shortcut
    dashboard-next-section
    dashboard-previous-line
    dashboard-refresh-buffer
    dashboard-previous-section
    dashboard-insert-startupify-lists
```

ricardoricho:
Hi, I've just submitted a PR to custom the function that I use with [perp-projectile](https://github.com/bbatsov/persp-projectile).
While I was testing I found `(setq projectile-completion-system 'ivy)` tells projectile always use ivy.

jsigman:
@ricardoricho When I tested this with dashboard it doesn't do anything when I select a project. Thanks though

jacksonbenete:
@jsigman , @ricardoricho solution worked for me.

![image](https://user-images.githubusercontent.com/2954722/91771434-625d9a00-ebb9-11ea-9328-43d3afc7381f.png)

Maybe Counsel needs any special call/require that you're not using?
I will show you the relevant conf in my init.el, maybe it will help:

```lisp
;; Ivy/Counsel
(ivy-mode 1)
(setq ivy-use-virtual-buffers t)
(setq enable-recursive-minibuffers t)

;; Ido
(require 'ido)
[...]

PR #232, opened 2020-06-27
Define custom function to switch to a project.

ricardoricho:
Add `dashboard-projectile-switch-function` to custom variables, a function use when switching to projectile projects.

By default is `projectile-switch-project-by-name` but it can be change to `counsel-projectile-switch-project-by-name`, in this case to solve #215 and in my case `projectile-persp-switch-project`.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -669,39 +669,58 @@ WIDGET-PARAMS are passed to the \"widget-create\" function."
 ;;
 ;; Projects
 ;;
+(defcustom dashboard-projects-switch-function
+  nil
+  "Custom function to switch to projects from dashboard."
+  :type 'function
+  :group 'dashboard)
+
 (defun dashboard-insert-projects (list-size)
   "Add the list of LIST-SIZE items of projects."
+  (dashboard-insert-section
+   "Projects:"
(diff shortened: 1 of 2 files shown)


jcs090218:
@ricardoricho Can you resolve the conflict for us? Thanks!

jcs090218 (approved):

dashboard-widgets.el
```diff
@@ -669,39 +669,59 @@ WIDGET-PARAMS are passed to the \"widget-create\" function."
...
+over custom backends."
+  (or dashboard-projects-switch-function
+      (cond
+       ((eq dashboard-projects-backend 'projectile)
+        'projectile-switch-project-by-name)
+       ((eq dashboard-projects-backend 'project-el)
+        (lambda (project)
+          "This function is used to switch to `PROJECT'."
+          (default-directory project)
+          (project-current-inhibit-prompt t)
```
jcs090218, line 718: I think there is something wrong with these lines? `default-directory` and `project-current-inhibit-prompt` are variables and not functions. 😕

ricardoricho: Thanks, I fixed it.

jcs090218 (approved):

jcs090218:
LGTM, Thanks! ;)

PR #234, opened 2020-08-05
Add option to open projects with magit

KaratasFurkan:
I would like to start to work on a project with magit. Before this PR, i was opening a random file and then open magit. It works for me now.

This PR is similar to #232 

Maybe something like this on that PR would work same (did not test yet):
```lisp
(setq dashboard-projectile-switch-function
      (lambda (arg) (let ((magit-display-buffer-function
                           'magit-display-buffer-same-window-except-diff-v1))
                      (projectile-vc arg))))
```

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -250,6 +250,12 @@ If nil it is disabled.  Possible values for list-type are:
   :type  '(repeat (alist :key-type symbol :value-type string))
   :group 'dashboard)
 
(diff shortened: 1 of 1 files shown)


KaratasFurkan:
(setq projectile-switch-project-action (lambda ()
                                         (let ((magit-display-buffer-function
                                                'magit-display-buffer-same-window-except-diff-v1))
                                           (magit))))

is projectile wide approach.

PR #238, opened 2020-09-04
Agenda refactor

ricardoricho:
Commits to refactor the agenda widget.
It show all agenda entries, expose `org-map-entries` `MATCH` parameter.
Allow time format.

### DIFF
--- a/README.org
+++ b/README.org
@@ -176,6 +176,15 @@ To customize which categories from the agenda items should be visible in the das
 (setq dashboard-org-agenda-categories '("Tasks" "Appointments"))
  #+END_SRC
 
+To have an extra filter, ~MATCH~ parameter is exposed as
+~dashboard-match-agenda-entry~ variable, by default is ~nil~
+#+begin_quote
+‘MATCH’ is a tags/property/TODO match. Org iterates only matched
+headlines. Org iterates over all headlines when MATCH is nil or t.
+#+end_quote
+
+See [[https://www.gnu.org/software/emacs/manual/html_node/org/Using-the-mapping-API.html][Org Manual]] for more information.
+
 ** Faces
 
 It is possible to customize Dashboard's appearance using the following faces:
--- a/README.org
+++ b/README.org
@@ -163,9 +163,9 @@ To customize it and customize its icon;
 (add-to-list 'dashboard-items '(agenda) t)
 #+END_SRC
 
-To show agenda for the upcoming seven days set the variable ~show-week-agenda-p~ to ~t~.
+To show agenda for the upcoming seven days set the variable ~dashboard-week-agenda~ to ~t~.
 #+BEGIN_SRC elisp
-(setq show-week-agenda-p t)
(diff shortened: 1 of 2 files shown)


jcs090218:
Sorry for the late reply!

@ricardoricho Can you resolve the conflict for us? Thank you!

ricardoricho:
Hi @jcs090218 I think is ready.

README.org
```diff
@@ -163,9 +163,9 @@ To customize it and customize its icon;
 (add-to-list 'dashboard-items '(agenda) t)
 #+END_SRC
 
-To show agenda for the upcoming seven days set the variable ~show-week-agenda-p~ to ~t~.
+To show agenda for the upcoming seven days set the variable ~dashboard-week-agenda~ to ~t~.
 #+BEGIN_SRC elisp
-(setq show-week-agenda-p t)
+(setq dashboard-week-agenda t)
```
jcs090218, line 168: May I ask, why we are specifying another variable but we could just use `show-week-agenda-p`? 😕

jcs090218, line 168: May I ask, why we are specifying another variable but we could just use `show-week-agenda-p`? 😕

dashboard-widgets.el
```diff
@@ -664,74 +669,98 @@ WIDGET-PARAMS are passed to the \"widget-create\" function."
...
-date part is considered."
-  (unless due-date
-    (setq due-date (current-time)))
-  (setq due-date (time-add due-date 86400))
-  (let* ((gregorian-date (dashboard-timestamp-to-gregorian-date timestamp))
-         (gregorian-due-date (dashboard-timestamp-to-gregorian-date due-date)))
-    (calendar-date-compare (list gregorian-date)
-                           (list gregorian-due-date))))
+(defcustom dashboard-week-agenda
+  t
```
jcs090218, line 673: I think for this kind of little code, we can just put `t` at the end of the variable name. Thanks!

Also for,

* `dashboard-week-agenda`
* `dashboard-agenda-time-string-format`
* `dashboard-match-agenda-entry`

jcs090218 (commented):
Sorry for the late reply! Some comments for this PR.

README.org
ricardoricho: It's a rename of the variable to match conventions:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html
In this case to let all known that `show-week-agenda-p` belongs to `dashboard` so it's `dashboard-week-agenda-p` but the suffix `-p` could be avoided since it's a boolean variable, that's `dashboard-week-agenda`

jcs090218: Got it! My first thought is that `show-week-agenda-p` is an external variable that belongs to `org-agenda` or `calendar`. I just realized it belongs to this package itself. Thanks for the confirmation!

jcs090218 (approved):

jcs090218:
Tested, and it works! Merged! Thank you!

PR #247, opened 2020-10-29
Move screenshot to new location.

jcs090218:
Minor, just moving `screenshot.png` under `etc` folder. I prefer not to mess image file with code.


JesusMtnez (approved):

JesusMtnez:
Missing CI here too. I do not worry in this case because there is no code change, but worth mentioning

jcs090218:
Merged this for now.

> Missing CI here too. I do not worry in this case because there is no code change, but worth mentioning

hmm... I don't have much experience with `circleci`; I think it should be a mis-setting from their CI tool? 😕

PR #256, opened 2020-11-09
Fixed compile warnings.

jcs090218:
Fixed the compile warnings during installation.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -49,7 +49,7 @@
 (declare-function org-map-entries "ext:org.el")
 (declare-function org-outline-level "ext:org.el")
 (declare-function org-today "ext:org.el")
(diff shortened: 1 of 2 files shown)


jcs090218:
Interesting... The lint failed because of this,

```
dashboard-widgets.el:0:0 (check-declare) (declare-function) function not found: "all-the-icons-fileicon" in file "ext:.emacs/elpa/all-the-icons-20200923.1339/all-the-icons.el"
dashboard-widgets.el FAIL
```

But when I tried install the package `all-the-icons`, it does appear in the file `all-the-icons.el`. 😕 Any idea?

JesusMtnez:
If you disabled those two `declare-funtion` everything works fine.. not sure what is happening :thinking:

PR #258, opened 2020-11-10
Add dashboard-after-initialize-hook to run after initialization

KaratasFurkan:
This could be useful to delay loading of some packages or any time consuming processes to end of the dashboard initialization.

Loading some big packages with this hook reduced my init time (init of dashboard screen actually) from 1.8s to 1.3s. Emacs does not respond before run-hooks finished of course but this feels much faster and I look to the dashboard for a sec anyway. So I think this is an improvement.

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -48,8 +48,10 @@
     map)
   "Keymap for dashboard mode.")
 
-(defvar dashboard-after-initialize-hook nil
-  "Hook that is run after dashboard buffer is initialized.")
+(defcustom dashboard-after-initialize-hook nil
+  "Hook that is run after dashboard buffer is initialized."
+  :group 'dashboard
+  :type 'hook)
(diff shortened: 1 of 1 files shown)


dashboard.el
```diff
@@ -48,6 +48,9 @@
     map)
   "Keymap for dashboard mode.")
 
+(defvar dashboard-after-initialize-hook nil
+  "Hook that is run after dashboard buffer is initialized.")
```
jcs090218, line 52: Can you change this to `defcustom`? And add `:group` and `:type` as parameters! Thanks!

KaratasFurkan: Done :ok_hand:

jcs090218 (approved):

jcs090218:
Merged! Thanks for the contribution!

PR #259, opened 2020-11-13
fix registers shortcut

sejgit:
very minor fix which prevented the registers "e" shortcut from working.


jcs090218 (approved):

jcs090218:
Thanks for the fix!

sejgit:
Welcome!  Thanks for all you do, been a loyal user for quite a while.

PR #261, opened 2020-11-14
Fixed compile warnings.

jcs090218:
This is the same as #256. I somehow mess up the base of the commitment so I just made another PR.

**Edit:** I also add a TODO for macro `all-the-icons-fileicon`. Not sure how to solve this yet. 😕


jcs090218:
The CI still fails. I think I will just leave this issue here for now. Still not sure how to solve these undeclare issue. 😕

dashboard.el
```diff
@@ -23,7 +23,6 @@
 ;;; Code:
 
 (require 'seq)
-(require 'page-break-lines)
```
jcs090218, line 26: `page-break-lines` is autoload. Don't need to require it on top.

jcs090218:
@JesusMtnez Can you give this a look? This PR should be ready to merge! Thanks!

JesusMtnez (approved):

jcs090218:
@JesusMtnez Thank you! ;)

PR #265, opened 2020-12-03
Use `project-switch-commands` for `project-el` backend of dashboard widget

rudi:
The project package has a command `project-switch-project` which lets the user choose first a project and then a command from a customizable list; e.g. find a file, open magit, start a shell, etc.

This pull request implements presenting the same list of commands from the project widget when `dashboard-projects-backend` is set to `project-el`, instead of always prompting for a file to open.


jcs090218:
Sorry, I think there are conflict after this PR #232. Can you resolve it for us? Thank you!

rudi:
I think after #232 this pull request might not be necessary anymore since the user can modify the behavior themselves.  I created #271 that documents how to use `dashboard-projectile-switch-function` and fix the customize widget of that variable.

jcs090218:
Thanks for explaining it! I thought this PR was different to #232. 😅 

Thanks for the contribution! I appreciated! ;)

rudi:
It does something different than #232, but just replaces one hard-coded behavior with another one.  #232 enables the user to choose the behavior themselves, so I think having the two project backends behave in the same way by default (i.e., prompt which file to open) is better overall for the project.

Thanks for working on dashboard! It's my default buffer for each new tab, and very useful overall.

PR #268, opened 2020-12-07
Add custom face for no items

jcs090218:
This patch added a custom face for displaying `--- No items ---`. I didn't bother to add the documentation here because I don't think this is the major feature. 

---

@JesusMtnez Let me know what you think! Thanks!


JesusMtnez (approved):
:shipit:

Issue #266, opened 2020-12-04
the dashboard showing aganda shows all TODO's 


algon-ns:
My Dashboard shows entries from next year, it seems like the filter do not work fdate/time not i US format.
![emacs-dash](https://user-images.githubusercontent.com/58687687/101221354-d2cbf980-3687-11eb-8c50-61b45485c44b.png)

jcs090218:
I think you are talking about the PR #238? If you want to change the time string try change the variable `dashboard-agenda-time-string-format`. 😕

---

Thanks for reporting this issue to us! 😄

algon-ns:
Sorry, 
the time format is OK (%Y-%m-%d)
I changed (show-....) to (dashboard....)
and I'm running emacs version 28.0.50, and the problem persist

and I didnot understand the answer for #238

jcs090218:
> and I didnot understand the answer for #238

No worries! I was trying to understand the issue here. I guess you are talking about the filter?

I think PR #238 has changed the filtering process to `dashboard-match-agenda-entry`. May be try tweaking the variable? 😕

algon-ns:
Thanks,

the dashboard-match-agenda…æ is nil, any suggestions of other values?

It look like I do not get the latest version of dashboard, I’m using straight.el to get the dashboard, and
I have no ideas how to get  (straight-use package to process all the config parameters

*** Dashboard
:PROPERTIES:
:CUSTOM_ID: Dashboard
:END:
 [[https://github.com/emacs-dashboard/emacs-dashboard][Dashboard]] - A startup dashboard which provides certain information 
             about your recent Emacs activities.
We want a Dashboard, which demand Page Break Line 
#+begin_src emacs-lisp
(straight-use-package '(page-break-lines :type git :flavor melpa :host github :repo "purcell/page-break-lines"))
(setq page-break-lines-lighter " ⩵")
(global-page-break-lines-mode)

(use-package dashboard
  :straight t 
  :diminish
  :bind ("<f2>" . open-dashboard)
  :custom
  (dashboard-banner-logo-title "Open EMACS for a new world...")
  (dashboard-startup-banner "/Users/niels/.emacs.d/images/ohands.png")
  (dashboard-items '( (bookmarks . 6) (agenda . 6)))
  (initial-buffer-choice (lambda () (get-buffer dashboard-buffer-name)))
  (dashboard-set-heading-icons t)
  (dashboard-set-navigator t)
  (dashboard-navigator-buttons
[...]

PR #269, opened 2020-12-09
Bugfix dashboard-filter-agenda-by-time.

ricardoricho:
Solves #266
When schedule-time or deadline-time is nil but not both,
org-time-less-p treat them as current-date. Filter will skip the entry
unless it's not done and has schedule time less than due date or has a
deadline time less than due date.

There are other ways to express this condition
```ruby
(when (or (org-entry-is-done-p)
          (and (or (null schedule-time)
                   (org-time-less-p due-date schedule-time))
               (or (null deadline-time)
                   (org-time-less-p due-date deadline-time))))
  (point))
```
Or
```ruby
(cond
  ((org-entry-is-done-p) (point))
  ((and schedule-time
        (org-time-less-p schedule-time due-date)) nil)
  ((and deadline-time
        (org-time-less-p deadline-time due-date)) nil)
  (t (point)))
```

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -779,13 +779,12 @@ if returns a point."
   (let ((schedule-time (org-get-scheduled-time (point)))
         (deadline-time (org-get-deadline-time (point)))
         (due-date (dashboard-due-date-for-agenda)))
-    (if (or (org-entry-is-done-p)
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Merged! Thank you so much!

PR #278, opened 2020-12-20
Support for image transforms in newer Emacs

zevlg:
In `dashboard-insert-image-banner` check also for image transforms
functionality to scale images to `:max-width` / `:max-heigh`

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -60,18 +60,20 @@
 (defcustom dashboard-image-banner-max-height 0
   "Maximum height of banner image.
 
-This setting applies only if Emacs is compiled with Imagemagick
-support.  When value is non-zero the image banner will be resized
-to the specified height, with aspect ratio preserved."
+This setting applies only if Emacs supports image transforms or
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
I have tested and it works! I wonder what does this patch does? Would it be necessary to check for transforming the image? Appreciate if you could elaborate it!

Anyway, thanks for the contribution! 😄

JesusMtnez (approved):

zevlg:
> I have tested and it works! I wonder what does this patch does? Would it be necessary to check for transforming the image? Appreciate if you could elaborate it!
> 
> Anyway, thanks for the contribution! 😄

Newer Emacs has image transformations functionality, so images could be scaled or rotated even if imagemagick is not compiled in (in newer Emacs imagemagick is disabled by default). This patch utilizes this image transforms functionality (if available) to scale Emacs logo

jcs090218:
Thanks for the elaboration! Merged!

Issue #264, opened 2020-11-28
Org Agenda not being highlighted on startup, but when saving the corresponding org file


sailikk:
1. I open up Emacs, and notice that I don't see the usual highlighting of the keyword NEXT or the text "Pay Credit Card" like they appear in my `school.org` file usually:
![image](https://user-images.githubusercontent.com/46624965/100518467-a253ff00-315f-11eb-91a6-65028ec3f40e.png)
2. (optional): Try to refresh the buffer a few times with `dashboard-refresh-buffer`. Note that nothing changes.
3. I press `a` to get to the agenda section, and point is immediately teleported to the "Pay Credit Card" heading shown here.
4. I get to `school.org` which opens up as a split window on the right
5. I immediately do `delete-window` (bound to `C-x 0` usually)
6. I am back to the `*dashboard*` buffer, and see this (the org heading is now highlighted)
![image](https://user-images.githubusercontent.com/46624965/100518462-9ec07800-315f-11eb-8e01-821a8713493d.png)

I would like the org heading to be highlighted as usual when I start up. Would this be a bug?

Edit: Here is also my `use-package dashboard` block:
```elisp
(use-package dashboard
  :init
  (add-hook 'after-init-hook 'dashboard-refresh-buffer)
  :config
  (setq dashboard-startup-banner 'logo)
  (dashboard-setup-startup-hook))
```

jcs090218:
Are you talking about something changes after the recent commits? Can you tell me what version of dashboard are you using? I think the issue appears after this PR here #238?

sailikk:
Sorry for late reply..

My elpa has a directory named dashboard-20201114.928, which should be the most recent MELPA version, right? New to Emacs, so please correct me if I have some misunderstanding. Oh, I think this is somewhat outdated. According to https://melpa.org/#/dashboard 20201207.1445 should be the most recent. What would be a good way to update it for someone using `use-package`? I will try to look this up as well.

I just started using dashboard a couple weeks ago. Also, I tried putting the `use-package dashboard` block as the last thing in my `init.el`, but nothing changes.

sailikk:
Okay, I figured out how to update my elpa directory packages, and everything is up to date now. The same thing is still happening with dashboard unfortunately

jcs090218:
> Regardless, 11/14 seems to be after the PR you mentioned was merged, so would this be a problem?

[...]

PR #281, opened 2020-12-27
Agenda highlight headlines

ricardoricho:
Introduces `dashboard-agenda-highlight-headlines` as a Boolean to
highlight agenda headlines using org default faces.

Headlines are highlighted by setting the corresponding
`face` text-property to each part of the headline. Issue #264 shows
that headlines are highlighted only after visit the org buffer and
when org buffer is killed headlines looses their faces.

I couldn't found the function that sets the faces, so now
faces are assigned to the TODO and HEADING parts only if they don't
already had a face.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -953,23 +953,16 @@ is todays date format."
     (dashboard-agenda--set-agenda-headline-face item)
     (list item loc file)))
 
-(defcustom dashboard-agenda-highligth-headlines t
-  "Ensure agenda headlines are highlighted."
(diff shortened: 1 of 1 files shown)


jcs090218:
Thanks for fixing this bug for us.

I wonder maybe we can remove variable `dashboard-agenda-highligth-headlines` and just highlight the headlines as the default behaviour? So user can have less variables to customize.

jcs090218 (approved):

jcs090218:
LGTM, thanks for making this package better! ;)

Issue #184, opened 2019-08-22
Mouse click errors


paxperscientiam:
Hello, I'm trying out emacs-dashboard. It works well in Emacs for Mac, but fails in dumb terminals. Basically, it seems like mouse clicking is broken.

![ezgif com-gif-maker](https://user-images.githubusercontent.com/7539871/63524854-b6c03b00-c4ca-11e9-9020-675b67bdf96d.gif)

jcs090218:
What terminal are you using? Just the default terminal from Mac? Might need more information on this.

paxperscientiam:
@jcs090218 iTerm2 on macos

JesusMtnez:
I think the problem is not related to `emacs-dashboard`. I do not use MacOS, so I could not assist you here.

Did you find a way to fix it @paxperscientiam ?

jcs090218:
Yeah, I cannot reproduce this on my Windows machine. I agree with @JesusMtnez that this may cause by the terminal you are using. 😕

zcaudate:
I'm having the same issue with the letters. Also on macOS.

zcaudate:
I've used Terminal.app, Term2 and Alacritty and they all do the same thing. I'm wondering if there was a way to disable the mouse

zcaudate:
I figured out how to disable it:

```
(define-key dashboard-mode-map [down-mouse-1] nil)
```

a work around is to click using the mouse and press RET on the file you want.

paxperscientiam:
@zcaudate  that indeed works. Just submitted PR #282 using your find.

jcs090218:
Do we even need this line in the first place? I think without this line the dashboard works functionally without causing issue.

https://github.com/emacs-dashboard/emacs-dashboard/blob/ef51ad205c6f21e80cb1e4b2ab2d0547675ea206/dashboard.el#L44

paxperscientiam:
My assumption is "yes." If not explicitly set in mode-map, we can't be sure what will be bound to it ... though perhaps it's likely just to be the default value.

jcs090218:
> My assumption is "yes." If not explicitly set in mode-map, we can't be sure what will be bound to it ... though perhaps it's likely just to be the default value.

I don't think we need this since `widget-button` should act as button object in Emacs. Unless it bounds to something else that mouse no longer do it usual behaviour.

[...]

PR #282, opened 2020-12-28
bugfix: test for terminal type for mode-map

paxperscientiam:
This resolves issue #184. Thanks to @zcaudate for the find.

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -41,12 +41,15 @@
     (define-key map (kbd "C-i") 'widget-forward)
     (define-key map [backtab] 'widget-backward)
     (define-key map (kbd "RET") 'dashboard-return)
-    (define-key map [down-mouse-1] 'widget-button-click)
     (define-key map (kbd "g") #'dashboard-refresh-buffer)
     (define-key map (kbd "}") #'dashboard-next-section)
     (define-key map (kbd "{") #'dashboard-previous-section)
     map)
   "Keymap for dashboard mode.")
+
+(if (display-graphic-p)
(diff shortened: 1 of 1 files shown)


dashboard.el
```diff
@@ -41,13 +41,16 @@
...
-    (define-key map [down-mouse-1] 'widget-button-click)
     (define-key map (kbd "g") #'dashboard-refresh-buffer)
     (define-key map (kbd "}") #'dashboard-next-section)
     (define-key map (kbd "{") #'dashboard-previous-section)
     map)
   "Keymap for dashboard mode.")
 
+(if (display-graphic-p)
+    (define-key dashboard-mode-map [down-mouse-1] 'widget-button-click)
+  (define-key dashboard-mode-map [down-mouse-1] nil))
```
jcs090218, line 52: Can you put this in the line of `defvar`? Maybe like

```el
(define-key map [down-mouse-1] (if (display-graphic-p) .. ..))
```

paxperscientiam: After talking with some folks on freenode#emacs, it seems that, while possible to put a similar expression within the defvar expression, it would still require code outside the defvar expression.

Alternatively, could leave the defvar expression alone and follow it with this:
```lisp
(unless (display-graphic-p)
    (define-key dashboard-mode-map [down-mouse-1] nil))
```

jcs090218: I don't think we need this part anymore since we removed the line L44. The default value should just be nil.

paxperscientiam: yeah, i think so. i'll close this

PR #285, opened 2021-01-02
Revert and use mouse-1 instead

jcs090218:
The fix from #283 isn't perfect, revert and attempt for next fix.

Reopen issue #184.


jcs090218:
Solution works and confirmed by @paxperscientiam! Thanks!

JesusMtnez (approved):

paxperscientiam:
awesome :)

PR #290, opened 2021-01-25
Add flag for force refresh

jcs090218:
I prefer not to kill the buffer and just refresh it with a flag.


jcs090218:
@JesusMtnez Can you take a look at this? The CI failed because the unremoved CircleCI. Thanks! :)

JesusMtnez (approved):

jcs090218:
Thanks!

PR #300, opened 2021-03-05
Fix icon for week agenda items

KaratasFurkan:
Without this PR, if `dashboard-week-agenda` is `t`, items have "unknown file icon".


jcs090218 (approved):
Thanks for the patch!

Issue #51, opened 2017-09-01
emacs-dashboard adds '~/.emacs.d/recentf' and 'org-agenda-files' list to top of recent files in dashboard


vonHabsi:
When `org-agenda` is added to the dashboard the files in `org-agenda-files` are added to the **Recent Files** list in the dashboard as well, and so is `~/.emacs.d/recentf` itself.

I suspect that the order of execution may be a factor here. Is there a way to avoid it at startup, regardless of the stage, perhaps ignoring `~/.emacs.d/recentf` and  `org-agenda-files` contents when constructing the dashboard the first time?

rakanalh:
Hi @vonHabsi,

It happens because dashboard actually opens these files and looks inside of them to get their contents.

Can you maybe try setting your `recentf-exclude` to ignore these files?
```
(setq recentf-exclude '("/path/"))
```

vonHabsi:
Thanks - I will look into that. I am a bit of emacs newbie.

vonHabsi:
Does emacs have a function or a mode for opening a file which doesn't add it to the recentf list?

rakanalh:
Not that i am aware of and i don't think so. did the recentf-exclude not work?

jhenahan:
If anyone is looking for a specific snippet that'll work in the general case

    (setq recentf-exclude (org-agenda-files))

or, if your agenda files are symlinks (as may be the case if you sync your files with a service but symlink them to your home directory), this snippet is what I use (requires `dash.el` and `f.el`):

    (setq recentf-exclude (-map 'f-canonical (org-agenda-files))

rakanalh:
Fixed with #83

PR #301, opened 2021-03-06
Fix recent files show org agenda files issue

fieryzig:
Hi, I have been using dashboard for a long time. Thanks for your amazing work and all the contributors.

Recently, I meet the issue that Recent Files show the org-agenda files.

I think #83 don't fix issue #51 perfectly.

#83 only works when the Emacs starts up the first time .

It won't works in the following situations:

1. press `g`(call `dashboard-refresh-buffer`) after you `kill-buffer` all your agenda files.

The reason here is #83 use `origial-recentf-list` to restore the `recentf-list` after ALL dashboard-items HAVE been rendered.
```
save recentf-list -> do agenda -> do other things -> do recentf -> restore recentf-list
```
I think the right thing is
```
save -> do agenda -> restore -> do other things -> restore -> do recentf -> restore
```
So I make this PR.

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -218,6 +218,8 @@ Optional prefix ARG says how many lines to move; default is one line."
                           (cdr-safe (assoc el dashboard-item-generators))))
                     (add-to-list 'dashboard--section-starts (point))
                     (funcall item-generator list-size)
+                    (when recentf-is-on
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Make sense to me! Thanks for the patch! :D

Issue #310, opened 2021-04-27
`(void-function path)` when loading bookmarks (dashboard-insert-bookmarks)


nwg:
I was getting the error `(void-function path)` when trying to start up dashboard my first time.

The stack trace was:

Debugger entered--Lisp error: (void-function path)
  path("/path/to/my-e.file")
  #f(compiled-function (el) #<bytecode 0x28ff081c9111c07>)("e")
  mapc(#f(compiled-function (el) #<bytecode 0x28ff081c9111c07>) ("e" "init"))
  dashboard-insert-bookmarks(5)
  #f(compiled-function (els) #<bytecode 0x59e82ce97221a20>)((bookmarks . 5))
  mapc(#f(compiled-function (els) #<bytecode 0x59e82ce97221a20>) ((recents . 5) (bookmarks . 5) (agenda . 5)))
  dashboard-insert-startupify-lists()
  #f(compiled-function () #<bytecode 0x1e00188cb435>)()
  run-hooks(after-init-hook delayed-warnings-hook)
  command-line()
  normal-top-level()

(my bookmark had name "e")

I tracked it down to the if-let* in dashboard-widgets.el / dashboard-insert-bookmarks.

if-let* was not a known function, so the code is trying to evaluate the left-hand-side of the let-bindings as a function.

Adding `(require 'subr-x)` to top of dashboard-widgets.el solved the problem for me.

Is this something that should be incorporated into the project? Or is there some other way to solve this problem?

jcs090218:
Ah, yes. You are right! Sorry for being careless on recent changes...😖 

Will you willing to open a PR for this? 😄 So I can approve and merge it ASAP!

PR #311, opened 2021-04-27
Add a missing require for subr-x #310

nwg:
This fixes issue #310. The symptom was a bad function call of a let
binding, path, in an if-let*


jcs090218 (approved):

jcs090218:
Thank you!

Issue #160, opened 2019-06-26
page-break-lines should be an optional dependency


Luis-Henriquez-Perez:
I don't really understand why this package is a _required_ dependency. It seems like it should be quite possible to display the content of the dashboard without page breaks. Or even, if you insist on some delineation you could have this:  `---------------------------------` .

JesusMtnez:
I agree that `page-break-lines` could be made an optional dependency. Right now I do not have plenty of time to work on something like that. But feel free work on that and open a PR. I'll try to give feedback whenever I can :smile:

Luis-Henriquez-Perez:
Ok sounds good. 👍 I was already planning to PR some things once I get more familiar with the code.

Luis-Henriquez-Perez:
Hey, so I'm more familiar with the code now. Looks like it's going to be really easy to remove `page-break-lines` as a dependency. It's only used 2 times in the code.

The integration between dashboard and `page-break-lines` is so thin that it seems that the ideal solution, would be to just not have page-break-lines in dashboard at all and have the user do this if they want the integration.

```elisp
(leaf page-break-lines
  :hook (dashboard-mode-hook . page-break-lines-mode))

(leaf dashboard
  :ensure t
  :require t
  :setp
  (dashboard-page-separator . "\n\f\n")
  (dashboard-center-content . t))
```
The benefit of doing this is that it's natural and simple. In general I think packages should not talk about one another unless one package is fundamental to the use of the other. If you choose not to go about it this way, then we'd need to add some sort of clunky option like `dashboard-use-page-break-lines` and add more if statements to the code.

On a related note, when I was reading `dashboard-insert-startupify-lists` I noticed the code below. 

```elisp
(when dashboard-center-content
            (goto-char (car (last dashboard--section-starts)))
            (let ((margin (floor (/ (max (- (window-width) max-line-length) 0)  2))))
              (while (not (eobp))
                (and (not (eq ? (char-after)))
                     (insert (make-string margin ?\ )))
                (forward-line 1))))
```
The point of it seems to be not to insert the margin before a page break. Right now it will only work if the page break is the default one `^L`, but if the user provides a custom page break it will insert the margin and the user's page-break will be really off (this happened to me). I think this code should be generalized (perhaps using `string-match-p` or something similar to check whether the text matches `dashboard-page-separator`).

JesusMtnez:
[...]

PR #315, opened 2021-07-12
Optional page-break-lines.

ricardoricho:
Hi, this is for #160

As @Luis-Henriquez-Perez said to keep the lines of page-break-lines it's needed to set `dashboard-page-separator` to "\n\f\n" and have installed page-break-lines (the hook is not needed).


jcs090218 (approved):

jcs090218:
@JesusMtnez What do you think?

jcs090218:
I think this should be good. Thanks for the changes! :)

PR #316, opened 2021-07-16
Allow section headings to be customized

### LINKED ISSUE #305: Customizable headlines in the widget
As I dont have english as my language, I would like to have the Headlines (like Bookmarks: og Projects) being made custormizable by defining them in a defvar/defcustom or the like.

BSaifullah:
Closes #305 

Section headings can now be given custom names through `dashboard-item-names'.
Since this is done through overlays it doesn't break any function that expects
the headings to have their default names.

It would arguably be more elegant to adapt all of the aforementioned cases to
use a system that can handle any name, but the work necessary would be a lot
more extensive than this, and I don't know how much of the codebase one would
want to revamp to accommodate this single feature.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -262,6 +262,17 @@ value is nil, that item's shortcut is disbaled.  See
   :type '(repeat (alist :key-type symbol :value-type string))
   :group 'dashboard)
 
+(defcustom dashboard-item-names nil
+  "Association list of item heading names.
+When an item is nil or not present, the default name is used.
+Will be of the form `(default-name . new-name)'.
+Possible values for default-name are:
+\"Recent Files:\" \"Bookmarks:\" \"Agenda for today:\",
+\"Agenda for the coming week:\" \"Registers:\" \"Projects:\"."
+  :type '(alist :key-type string :value-type string)
+  :options '("Recent Files:" "Bookmarks:" "Agenda for today:"
+             "Agenda for the coming week:" "Registers:" "Projects:"))
+
 (defcustom dashboard-items-default-length 20
   "Length used for startup lists with otherwise unspecified bounds.
 Set to nil for unbounded."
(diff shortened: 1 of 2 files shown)


dashboard-widgets.el
```diff
@@ -438,6 +450,8 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
     (insert " "))
 
   (insert (propertize heading 'face 'dashboard-heading))
+  (overlay-put (make-overlay (- (point) (length heading)) (point))
+               'display (or (cdr (assoc heading dashboard-item-names)) heading))
```
jcs090218, line 454: You might want to preserve the face, and this does not work when `dashboard-center-content` is set to `t`. Make sure to pass in `t` to make-overlay's fourth argument FRONT-ADVANCE so the insertion will be covered.

hence,

```el
;; Some comment here...
(let ((ov (make-overlay (- (point) (length heading)) (point) nil t)))
    (overlay-put ov 'display (or (cdr (assoc heading dashboard-item-names)) heading))
    (overlay-put ov 'face 'dashboard-heading))
```

Can you also add some comment about the changes?

jcs090218 (commented):
Can you make these changes for us? Thanks!

dashboard-widgets.el
BSaifullah: @jcs090218 ,

I applied your changes and added a comment.  Thank you for the feedback.

>You might want to preserve the face
During my (admittedly brief) tests I found that the original face was already
preserved since the overlay didn't set one to override it.
But I agree, it's better to explicitly apply the face again just to make sure.

jcs090218: > During my (admittedly brief) tests I found that the original face was already
preserved

I think  you are right, but it does not preserved on my machine. This seems strange. 😕

jcs090218:
Can you fix the CI for us?

And I forgot to mentioned, can you add document to our README file so the user would know this nice new feature? 😁 Thanks!

BSaifullah:
Done :-)

jcs090218 (approved):
Thank you for contribute to emacs-dashboard. 😁

PR #317, opened 2021-07-18
Update document for optional dependency

jcs090218:
`page-break-lines` became optional dependency after PR #315.

Update the README file.

---

By the way, @JesusMtnez can you set me as a org owner? or grant me the push permission to this repo? I think this would be a lot easier for me to maintain it! Hope I am qualify enough for the permission! 😅

### DIFF
--- a/README.org
+++ b/README.org
@@ -21,7 +21,7 @@ An extensible emacs startup screen showing you what's most important.
 * Dependencies
 You will need the following packages which are all available on Melpa:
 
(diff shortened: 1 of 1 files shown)


JesusMtnez (approved):
I will take a look at your permissions, you are right now the most active maintainer, so you should have those permissions.

jcs090218:
That will be great! Thank you! 😁

PR #330, opened 2021-09-28
Fix init-info reporting wrong `emacs-init-time'

andrewhr:
This patch fixes an issue where the init info doesn't show the correct value of `emacs-init-time`.

The cause is due the `defcustom` default value being computed at whatever time this package is loaded, which not necessarily represent the final startup time. As example, on my local use-package based setup it reports something around 0.6 seconds, while manually calling `M-x emacs-init-time` does say 2.5 seconds.

The fix is wrapping the default message inside a lambda, to defer computing until render time. This change should keep old behavior, in case any user override `dashboard-init-info` with some fixed string.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -158,16 +158,17 @@ Example:
   :group 'dashboard)
 
 (defcustom dashboard-init-info
-  (let ((package-count 0) (time (emacs-init-time)))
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Thanks!

Issue #326, opened 2021-09-07
Agenda sort entries by scheduled date


ghost:
The entries are being apparently sorted by the last modified file to the first, but I would like to sort them by their scheduled date. Is it a bug, or just misconfiguration?

mlemosf:
I'm having the same problem, the entries are sorted on the agenda view but not on the dashboard.

jcs090218:
You can now use variable `dashboard-agenda-sort-strategy` to sort.

PR #335, opened 2021-10-24
Dashboard agenda sort

ricardoricho:
Hi, this is for #326
In org there is [org-agenda-sorting-strategy](https://orgmode.org/worg/doc.html#org-agenda-sorting-strategy) to sort agenda entries.
This PR introduces `dashboard-agenda-sort-strategy` to define the list of strategies for sorting the agenda.
For example, in the case of sorting by time (scheduled) `(setq dashboard-agenda-sort-strategy '(time-up))` will put the entries with time indications first (early first).


jcs090218 (approved):
Thanks!

PR #340, opened 2021-12-19
Filter out archived headings from agenda items

logc:
This change removes archived items, i.e. headings under a tree with the ARCHIVE tag, from the agenda shown on the dashboard. Tested locally on my Emacs setup.

Please let me know if there is something missing from this PR (documentation changes, tests, etc ...). I hope this contribution is useful to others apart from me: showing archived items should not happen in an agenda view. I guess this happens in the dashboard because it is not a standard org-agenda view (?)


jcs090218:
The CI does not pass. Seems like the function `org-in-archived-heading-p` does not defined in `org-mode`? Is the function only exists in the newer `org-mode`? 😕

jcs090218 (approved):

logc:
I have tried out the change on the following versions, quite recent because recently re-installed:

- GNU Emacs 27.2 (build 1, x86_64-apple-darwin19.6.0, Carbon Version 162 AppKit 1894.6) of 2021-11-24
- Org mode version 9.5.1 (9.5.1-g895e0b @ ...)

jcs090218:
It does make sense. Not sure if we should add a custom variable for this?

I will merge this for now! Thanks for the contribution! 👍

PR #343, opened 2021-12-21
Simplify with cl-case instead of cond

jcs090218:


dashboard-widgets.el
```diff
@@ -1193,12 +1193,11 @@ found for the strategy it uses nil predicate."
...
-   ((eq 'time-up strategy) 'org-time-less-p)
-   ((eq 'time-down strategy) (lambda (a b) (org-time-less-p b a)))
-   ((eq 'todo-state-up strategy) '>)
-   ((eq 'todo-state-down strategy) '<)
-   (t nil)))
+  (cl-case strategy
+    (`time-up 'org-time-less-p)
+    (`time-down (lambda (a b) (org-time-less-p b a)))
+    (`todo-state-up '>)
+    (`todo-state-down '<)))
```
jcs090218, line 1200: We don't need `(t nil)` since it will just return `nil` when nothing matches.

PR #354, opened 2022-01-28
correct straight.el package count

cburroughs:
In emacs the "size" of as hash table refers to how big the table
itself is.  In other words, its capacity.  "count" is for "the actual
number of entries".


jcs090218 (approved):

jcs090218:
Thanks!

PR #350, opened 2022-01-20
Custom dashboard-agenda-prefix-format.

ricardoricho:
Hi, this is for #339. 

In that case, for a 'LONGLONGCATEGORY' with length 16 the `dashboard-agenda-prefix-format` could be `" %i %16c %-10s "`. I hope the comments on the README file are clear enough. 
Thanks.

PR #355, opened 2022-01-29
Fix #353 space between todo and headline.

### LINKED ISSUE #353: No space between TODO and todo item
In the agenda items list there are no spaces between the TODO and the todo item. It looks like

...   TODOthis has to be done
...   TODOnext item to be done

I'm using Emacs 29.0.50, everything else is from MELPA and is up-to-date.

ricardoricho:
I also modify some code as suggested in #350 

I change `add-to-list` with another `if-let` trying to keep the list `org-agenda-prefix-format` as clean as possible and prevent to keep adding a `'dasboard-agenda` entry each time `dashboard-agenda-prefix-format` change.


jcs090218 (approved):

jcs090218:
Perfect! Thanks for the changes! :)

PR #358, opened 2022-02-08
Add Lsp Session folders widget

gtejedap:
Add Lsp Session folders widget


jcs090218:
I am not sure about this. Shouldn't this be in a separate package? Maybe a package name `dashboard-lsp`, `dashboard-lsp-folders`, etc? 😕

gtejedap:
Yes, you're right. Probably the best is move it into a separated package.
I'll be sharing to you the link of the repository once is done (in case you want to include it as reference in the documentation)

PR #359, opened 2022-02-08
Lazy load it's module

jcs090218:
To avoid force loading certain modules to increase startup time.

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -61,12 +61,10 @@
   :syntax-table nil
   :abbrev-table nil
   (buffer-disable-undo)
-  (whitespace-mode -1)
-  (linum-mode -1)
-  (when (>= emacs-major-version 26)
-    (display-line-numbers-mode -1))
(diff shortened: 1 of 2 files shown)


dashboard.el
```diff
@@ -61,12 +61,10 @@
...
-  (whitespace-mode -1)
-  (linum-mode -1)
-  (when (>= emacs-major-version 26)
-    (display-line-numbers-mode -1))
-  (when (require 'page-break-lines nil t)
-    (page-break-lines-mode 1))
+  (when (featurep 'whitespace) (whitespace-mode -1))
+  (when (featurep 'linum) (linum-mode -1))
+  (when (featurep 'display-line-numbers) (display-line-numbers-mode -1))
+  (when (featurep 'page-break-lines) (page-break-lines-mode 1))
```
jcs090218, line 67: Even calling `(minor-mode -1)` will load it's module. Avoid doing that can increase startup time.

Issue #154, opened 2019-06-19
Delete items?


jcs090218:
Is there a way to delete the currently existing item? Like removing a file from the `Recent Files: (r)` list.

JesusMtnez:
I suppose you can achieve that if `recentf` has a method to delete items from its list. I know `projectile` has a method to do that.

Let me know if this answers your question to close the issue

jcs090218:
I was asking if `dashboard` would want to achieve something like this! It is okay if you don't want this feature in this package! :D I guess this issue is a new feature request? :/

JesusMtnez:
@jcs090218 Of course it would be a great feature. Feel free to open a PR to add it :+1:

PR #366, opened 2022-03-26
Add functionality to delete items

jcs090218:
For #154


jcs090218:
Seems like the CI is false-positive. I will come back, and fix CI later on.

PR #372, opened 2022-04-09
Agenda remove item

ricardoricho:
Hi, this is the implementation of `dashboard-remove-item-agenda`, it calls `org-todo`  with the entry at point and if the `todo-state` is changed to `DONE` (or any filter state) then the entry is removed for the dashboard.


jcs090218 (approved):
Awesome! I wasn't have time to implement for the agenda but now we have it! 😄

PR #378, opened 2022-05-26
Make sure dashboard-resize-on-hook run at the very end of window-size-change-functions

cxa:


jcs090218:
Thanks for the changes!

I wonder what does this solved? Is there a benefit doing this? Maybe a brief explanation would be great! :)

cxa:
I'm also using `window-size-change-functions` to make my Emacs window shrink and center, to keep `dashboard-resize-on-hook` run after my window resizing customization, this hack is needed. for ref, I attach a screenshot without this hack:
Screen Shot 2022-05-26 at 14 40 18

jcs090218:
Couldn't you refresh dashboard buffer after the set up? Or may be re-order the `window-size-change-functions` to the desire order? 

Hope the information helps! 😅

cxa:
Thanks for your information! As I consult the `add-hook` doc, I found the optional third arg `depth` is very helpful for this situation.

jcs090218 (approved):
Awesome! I am not quite sure if this would cause any side effect, but by reading docstring it should be an ideal solution.

Let's try this out!

Issue #346, opened 2021-12-26
Add ace-link to dashboard


ddoherty03:
This is a wish:  optionally allow links to be opened with ace-links so that the feel is consistent with other link-based modes, such as info, etc.

ricardoricho:
Hi @ddoherty03 there is a new package https://github.com/emacs-dashboard/ace-link-dashboard

jcs090218:
I have added ace-link support to [JCS-ELPA](https://github.com/jcs-emacs/jcs-elpa), just in case if anyone want to use it through `package.el`. It would be nice if we eventually publish it to MELPA.

PR #380, opened 2022-06-14
Add ace-link-dashboard

ricardoricho:
Hi, this is for #346 
This doesn't depend on [ace-link](https://github.com/abo-abo/ace-link), just depends on [avy](https://github.com/abo-abo/avy) and still has to set a key: 
```emacs-lisp
(define-key dashboard-map (kbd "l") 'ace-link-dashboard)
```

### DIFF
--- a/ace-link-dashboard.el
+++ b/ace-link-dashboard.el
@@ -1,0 +1,44 @@
+;;; ace-link-dashboard.el -- Ace link for dashboard-mode
+;;; Commentary:
+;;; Code:
+(declare-function avy-with "ext:avy.el")
+(declare-function avy-process "ext:avy.el")
+(declare-function avy--style-fn "ext:avy.el")
+(declare-function avy-with "ext:avy.el")
+(declare-function widget-move "ext:wid-edit.el")
+(declare-function widget-at "ext:wid-edit.el")
+(declare-function widget-button-press "ext:wid-edit.el")
+
+(defvar avy-style nil "Originally defined in avy.el.")
+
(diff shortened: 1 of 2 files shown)


jcs090218:
Wouldn't this better to live as an external package? I think we should have majorly used features in this package and then have others live externally. This way would be more lightweight, and could save some startup time esp. for Windows.

ricardoricho:
Hi @jcs090218 I created https://github.com/ricardoricho/ace-link-dashboard 
Could you help me by reviewing it and release it as a package?
Maybe could be part of the [emacs-dashboard](https://github.com/emacs-dashboard) workspace

jcs090218:
Nice! Looks good to me! How would you like to do to further progress this? Do you want to publish it somewhere? MELPA?

You can move the repo to this org if you want to (not mandatory)!

ricardoricho:
Now it is on https://github.com/emacs-dashboard/ace-link-dashboard It would be nice to publish on MELPA, add a license, manage dependencies ([avy](https://github.com/abo-abo/avy)), some linter, etc. Maybe [eask](https://emacs-eask.github.io/)?

jcs090218:
I have opened up a PR in https://github.com/emacs-dashboard/ace-link-dashboard/pull/1. I think it should be good now if you want to publish it to MELPA.

Another option is [JCS-ELPA](https://github.com/jcs-emacs/jcs-elpa) (my own elpa), if you don't want to wait for MELPA's long review process and just want to use it right away.

Issue #379, opened 2022-06-09
When initialized, advance to the frst line below logo


mmahmoudian:
Thanks for this fantastic package. At the moment when the dashboard opens (e.g when `emacs` is executed), the point is at the beginning of the buffer, meaning right before the logo. This would make the user to always at least press a Tab to navigate to the first section. It would be nice if dashboard can already put the pointer at the first item of the first section.

![image](https://user-images.githubusercontent.com/390889/172812799-c7d44123-5906-4f54-8401-c8ca7a14c0b7.png)

mmahmoudian:
I wag going back and forth with [another issue](https://github.com/emacs-dashboard/emacs-dashboard/issues/384#issuecomment-1159202963) that I realized the `(dashboard-goto-recent-files)` does not exist. I ultimately landed on the following which practically does what I proposed here:

```elisp
(dashboard--goto-section-by-index 1)
(next-line)
```

This is not extremely elegant as ideally the cursor should not even be able to go to the line above "Recent Files", but still this approach satisfies the suggestion I made.

dian-yu-luo:
that doesn't work

mmahmoudian:
What is "that"? Do you mean the PR #387 ? Because it works just fine on vanilla Emacs as far as I can tell.

dian-yu-luo:
Merging is blocked,
so do I need to add it by myself?

jcs090218:
Sorry for the late reply!

I was confused by the first look, but I understand it now. I've reviewed the PR!

dian-yu-luo:
I am trying in this way , ` (add-hook 'dashboard-mode-hook 'dashboard-jump-to-recents)`
but it doesn't work when I swither buffer from other by `dashboard-refresh-buffer`. but it is useful if I refresh *dashboard*,why

mmahmoudian:
> it doesn't work when I swither buffer from other by `dashboard-refresh-buffer`. but it is useful if I refresh dashboard

If I don't make mistake, the hooks are only called when the buffer enters the mode, and not when you switch buffers. I personally don't know any hook that gets triggered when switching buffers

jeroenvanriel:
> I am trying in this way , ` (add-hook 'dashboard-mode-hook 'dashboard-jump-to-recents)`

This did not work for me, I had to use `(add-hook 'dashboard-after-initialize-hook 'dashboard-jump-to-recents)` instead.

PR #387, opened 2022-07-07
Move point to the first section

mmahmoudian:
This addresses #379

I'm very new to emacs and elisp, therefore although this change is very small, but still it needs to be checked.

If this works out and is up to your standard, I will create another PR for another issue that I have managed to solve on my end.

Thanks.

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -458,7 +458,8 @@ Dashboard."
                                  (dashboard-insert-startupify-lists)))
     (add-hook 'emacs-startup-hook (lambda ()
                                     (switch-to-buffer dashboard-buffer-name)
-                                    (goto-char (point-min))
+                                    (dashboard--goto-section-by-index 1)
+                                    (next-line)
                                     (redisplay)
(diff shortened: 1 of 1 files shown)


jcs090218:
I would suggest to add a defcustom variable for this behavior, WDYT?

ricardoricho:
Hi, I think the solution in [this comment](https://github.com/emacs-dashboard/emacs-dashboard/issues/379#issuecomment-1657217011) is the right one:
```emacs-lisp
(add-hook 'dashboard-after-initialize-hook 'dashboard-jump-to-recents)
```
If you want to move to the first item in the first section then:
```emacs-lisp
(add-hook 'dashboard-after-initialize-hook 
  (lambda ()
      (dashboard-goto-section-by-index 1)
	  (widget-forward 1)))
```
Maybe we need another hooks that run before and after insert the startupify
list, because `dashboard-after-initialize-hook` runs inside
`dashboard-initialize` and we don't always use that function

PR #390, opened 2022-07-15
fix: add projectile support

dian-yu-luo:


jcs090218:
We might need to forward declare the function on top with `declare-function`. Would you mind adding that? Thanks!

jcs090218 (approved):

PR #391, opened 2022-07-17
Use `string-width` to compute the width of the dashboard footer

Elilif:
`length` cannot compute the width of a string on display, so the footer may not be at the center of the screen  when its language is not English,


jcs090218 (approved):

jcs090218:
LGTM! Thanks!

PR #398, opened 2022-08-28
Fixed issue wherein all previous entries showed.

anerisgreat:
Current function dashboard-filter-agenda-by-time allows all past agenda items to show.

This means all past items are passed.

This filters agenda items, and passes them only if an associated date is after current date.

This makes sure the items are relevant to the current time period, making sure they either start or end in the current week and not at some point in the past.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -1204,7 +1204,8 @@ point."
   (let ((scheduled-time (org-get-scheduled-time (point)))
         (deadline-time (org-get-deadline-time (point)))
         (entry-timestamp (dashboard-agenda--entry-timestamp (point)))
-        (due-date (dashboard-due-date-for-agenda)))
+        (due-date (dashboard-due-date-for-agenda))
+        (now (current-time)))
     (unless (and (not (org-entry-is-done-p))
                  (not (org-in-archived-heading-p))
                  (or (and scheduled-time
@@ -1212,7 +1213,15 @@ point."
                      (and deadline-time
                           (org-time-less-p deadline-time due-date))
                      (and entry-timestamp
-                          (org-time-less-p entry-timestamp due-date))))
+                          (org-time-less-p entry-timestamp due-date)))
+                 (or
+                    (org-entry-is-todo-p)
+                    (and scheduled-time
+                          (org-time-less-p now scheduled-time))
+                    (and deadline-time
+                          (org-time-less-p now deadline-time))
+                    (and entry-timestamp
+                          (org-time-less-p now entry-timestamp))))
       (point))))
 
 (defun dashboard-filter-agenda-by-todo ()
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -1213,15 +1213,8 @@ point."
                      (and deadline-time
                           (org-time-less-p deadline-time due-date))
                      (and entry-timestamp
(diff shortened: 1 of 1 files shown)


jcs090218:
@ricardoricho Can you review this?

Also, Happy New Year! 🥳 🎉

ricardoricho:
@jcs090218 @anerisgreat Happy new year!!

I think this issue requires a little discussion. In my case, I found useful to show all entries even if they are overdue (or specially those). I mean, if I was supposed to finish a task in this week, but I didn't, it would be nice to show it next week until I marked as `DONE`.  It's not the case for the entries with just a timestamp.

Another issue is that we are filtering the entries with `(not (org-entry-is-done-p))` which is different from `(org-entry-is-todo-p)`, I prefer the latest so, maybe we also need to change that.

I propose this as the body of the function
```emacs-lisp
(let ((scheduled-time (org-get-scheduled-time (point)))
        (deadline-time (org-get-deadline-time (point)))
        (entry-timestamp (dashboard-agenda--entry-timestamp (point)))
        (due-date (dashboard-due-date-for-agenda))
        (now (current-time)))
    (unless (and (org-entry-is-todo-p)
                 (not (org-in-archived-heading-p))
                 (or (and scheduled-time
                          (org-time-less-p scheduled-time due-date))
                     (and deadline-time
                          (org-time-less-p deadline-time due-date))
                     (and entry-timestamp
                          (org-time-less-p now entry-timestamp)
                          (org-time-less-p entry-timestamp due-date))))
      (point)))
```
What do you think?

anerisgreat:
Happy new year to all!!

I agree with the first suggestion. Overdue items (if scheduled or deadlined)
should be shown. Other timestamps indicate simple date at which it
happened, and then we can filter them out.

Second suggestion - not so sure. `(not (org-entry-is-done-p))` is what
filters out entries that are already marked done, therefor, needs to be
kept in.

You don't necessarily want to filter out based on
`(org-entry-is-todo-p)`. My use-case for this is scheduling - I have
appointments written (synced from NextCloud in my case) and they appear
in the agenda. These are not todo entries, they happen whether I do them
or not (appointments/events/holidays/etc), yet they are org entries that
do appear as part of a greater schedule.

Would we like to possibly add a variable to allow a user whether only
TODO entries are to be displayed as part of the agenda?

Either way, thank you for your thoughtful consideration!

-Aner

anerisgreat:
I updated based on suggestions, but kept `(not (org-entry-is-done-p))` per rationale writen above.

I will also say that I take back my suggestion of adding variables - I think if people want more specific filtering, at this point, they can write their own custom filtering function (as is supported by the library).

I do think default behavior that entries with simple timestamp should only be shown if the time falls in relative area is a good idea, and also that it should also show not only TODO entries, which is why I think it is worthwhile changing the default filter-by-time method.

ricardoricho (approved):
Great

ricardoricho:
Hi @jcs090218, we are ready to merge this PR, could you do it? Does it need to be rebased, squashed or both?

jcs090218:
> we are ready to merge this PR, could you do it?

@ricardoricho I have given you the Write status for this repo. Can you merge PRs now? 🤔 

> Does it need to be rebased, squashed or both?

Yes, it's better to rebase it! ;)

PR #399, opened 2022-09-06
Support `Elpaca` in `dashboard-init-info`

Riyyi:
Add support for the number of packages from Elpaca.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -174,6 +174,8 @@ Example:
         (setq package-count (length package-activated-list)))
       (when (boundp 'straight--profile-cache)
         (setq package-count (+ (hash-table-count straight--profile-cache) package-count)))
+      (when (fboundp 'elpaca--queued)
+        (setq package-count (length (elpaca--queued))))
       (if (zerop package-count)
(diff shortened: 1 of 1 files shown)


dashboard-widgets.el
```diff
@@ -174,6 +174,8 @@ Example:
         (setq package-count (length package-activated-list)))
       (when (boundp 'straight--profile-cache)
         (setq package-count (+ (hash-table-count straight--profile-cache) package-count)))
+	  (when (fboundp 'elpaca--queued)
+		(setq package-count (length (elpaca--queued))))
```
jcs090218, line 178: Let's use spaces instead of tab here! And the indentation seems to be incorrect. Can you please fix that also? Thanks!

jcs090218 (approved):
LGTM, some nits.

dashboard-widgets.el
Riyyi: Whoops, I ignored the project's .dir-locals.el, my bad. The indentation is correct AFAIK, its the same as the two cases above.

jcs090218:
Thanks!

PR #407, opened 2022-11-05
Center banner with properties and combine text with image

### LINKED ISSUE #197: Resize hook affects all frames
When using dashboard and emacsclient, resizing an emacs frame triggers the resize hook and all frames, making some frames uncentered.

Before:

2019-11-11-131543_1920x1080_scrot

After:

2019-11-11-131552_1920x1080_scrot

### LINKED ISSUE #386: Center content before GUI loads
Currently, the dashboard is first shown left-aligned and then a short moment later the content snaps to the center, if configured. While not a functional problem, this is pretty jarring.

I was wondering whether this is expected behaviour in the first place (maybe there's something wrong with my config), and if so whether this is fixable.

I've tried moving different config settings to early-init.el, but I didn't manage to fix it.

### LINKED ISSUE #388: Better approach for emacsclient
The following information from the README is suboptimal:

> In addition to the above, configure `initial-buffer-choice` to show Dashboard in frames created with `emacsclient -c` as follows:
> 
> ```elisp
> (setq initial-buffer-choice (lambda () (get-buffer-create "*dashboard*")))
> ```

This is suboptimal because the first time after `emacs --bg-daemon` that a frame is created, it will not properly load the dashboard with all the graphics:

image

and the user have to reload the config to get it fixed. This is most probably because the `initial-buffer-choice` is run before any frame exists, and therefore everything is in text-only mode.

A better way would be to have

```elisp
(add-hook 'server-after-make-frame-hook (lambda()
    (switch-to-buffer dashboard-buffer-name)
    (dashboard-mode)
    (dashboard-insert-startupify-lists)
    (dashboard-refresh-buffer)))
```

which uses [a proper hook](https://www.gnu.org/software/emacs/manual/html_node/elisp/Creating-Frames.html#index-server_002dafter_002dmake_002dframe_002dhook) and also does not disturb the `emacsclient --tty --create-frame`. The content of the lambda is inspired from https://github.com/emacs-dashboard/emacs-dashboard/issues/236#issuecomment-711360926 

I suggest removing this section from README (or move it to some other section) and instead extend the `dashboard-setup-startup-hook` function to contain the hook above as well.

If you agree with this suggestion, I can create a PR.

Stebalien:
This patch contains two (entangled) fixes:

1. Center the banner with text properties.
2. Display both the image and text banner at the same time, letting the frame pick the right one for the display type.

The first change, centering with text properties, ensures that the text is smoothly centered without flickering, even if the dashboard is displayed in multiple windows at the same time (each window will center independently).

I'm inserting both the text and image banner at the same time to ensure that we always get the correct banner. Before this change, creating the dashboard buffer before creating a GUI frame would display a text banner.

Known issue: This patch still has to pick whether to center by the image or by the text banner. It does this at redisplay time using a display conditional, but it gets a bit confused when switching back and forth between a terminal frame and a GUI frame.

fixes #386
fixes #197
related #388 (icons still aren't handled)

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -152,9 +152,6 @@ preserved."
   (concat dashboard-banners-directory "logo.png")
   "Emacs banner image.")
 
-(defconst dashboard-banner-length 75
-  "Width of a banner.")
-
 (defcustom dashboard-banner-logo-title "Welcome to Emacs!"
   "Specify the startup banner."
   :type 'string
@@ -211,14 +208,14 @@ Default value is `official', it displays the Emacs logo.  `logo' displays Emacs
 alternative logo.  An integer value is the index of text banner.  A string
 value must be a path to a .PNG or .TXT file.  If the value is nil then no banner
 is displayed."
-  :type '(choice (const  :tag "offical"   official)
-                 (const  :tag "logo"      logo)
-                 (string :tag "a png or txt path"))
-  :group 'dashboard)
-
-(defcustom dashboard-buffer-last-width nil
-  "Previous width of dashboard-buffer."
-  :type  'integer
+  :type '(choice (const   :tag "no banner" nil)
+                 (const   :tag "offical"   official)
+                 (const   :tag "logo"      logo)
+                 (integer :tag "index of a text banner")
+                 (string  :tag "a path to an image or text banner")
+                 (cons    :tag "an image and text banner"
+                   (string :tag "image banner path")
+                   (string :tag "text banner path")))
   :group 'dashboard)
 
 (defcustom dashboard-item-generators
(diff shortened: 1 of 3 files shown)


jcs090218:
Cool, I think this patch would also resolve issue #197?

I am currently having a disability to work with my PC. Would someone like to test and review this patch? 

Just randomly pining some maintainers here, @mveytsman @ricardoricho

Stebalien:
> Cool, I think this patch would also resolve issue https://github.com/emacs-dashboard/emacs-dashboard/issues/197?

Yes, it should fix that as well (added to the list).

Stebalien:
It may also fix https://github.com/emacs-dashboard/emacs-dashboard/issues/292?

JesusMtnez:
I have enabled the CI for the PR, but I can not tested locally. I have reviewed the code and looks good to me. If it is OK for you, @jcs090218 @ricardoricho , we could merge it. Feel free to approve if you agree with it.

dashboard.el
```diff
@@ -388,66 +388,52 @@ Optional argument ARGS adviced function arguments."
...
+                      (let* ((el (or (car-safe els) els))
+                             (list-size
+                              (or (cdr-safe els)
+                                  dashboard-items-default-length))
+                             (item-generator
+                              (cdr-safe (assoc el dashboard-item-generators))))
+                        (push (point) dashboard--section-starts)
+                        (funcall item-generator list-size)
+                        (goto-char (point-max))
+                        (when recentf-is-on
```
ricardoricho, line 408: I think we don't need to restore the `recentf-list` to it's original value (also I think there is a typo to fix in `origial-recentf-list`) It's done at the end of the function.

dashboard.el
```diff
@@ -388,66 +388,52 @@ Optional argument ARGS adviced function arguments."
...
+              (insert dashboard-page-separator)
+              (save-excursion
+                (dolist (start dashboard--section-starts)
+                  (goto-char start)
+                  (insert dashboard-page-separator)))
+              (dashboard-insert-footer))
+            (goto-char (point-min))
+            (dashboard-mode))
+          (current-buffer))
+      (when recentf-is-on
```
ricardoricho, line 428: Here we are restoring the `recentf-list`

dashboard.el
```diff
@@ -388,66 +388,52 @@ Optional argument ARGS adviced function arguments."
...
-            (add-hook 'window-size-change-functions 'dashboard-resize-on-hook 100)
-            (dashboard-resize-on-hook)))
+    (prog1
+        (with-current-buffer (get-buffer-create dashboard-buffer-name)
+          (when (or dashboard-force-refresh (not (eq major-mode 'dashboard-mode)))
+            (let (buffer-read-only)
+              (erase-buffer)
+              (dashboard-insert-banner)
+              (setq dashboard--section-starts nil)
+              (mapc (lambda (els)
```
ricardoricho, line 398: Maybe we can named this `lambda` to `dashboard-items-list` or something more descriptive, I think is an important function and deserves a name 😄

dashboard-widgets.el
```diff
@@ -211,14 +208,14 @@ Default value is `official', it displays the Emacs logo.  `logo' displays Emacs
...
-
-(defcustom dashboard-buffer-last-width nil
-  "Previous width of dashboard-buffer."
-  :type  'integer
+  :type '(choice (const   :tag "no banner" nil)
+                 (const   :tag "offical"   official)
+                 (const   :tag "logo"      logo)
+                 (integer :tag "index of a text banner")
+                 (string  :tag "a path to an image or text banner")
+                 (cons    :tag "an image and text banner"
```
ricardoricho, line 216: This is a new option, Should we add an example on the `README` file?

ricardoricho (commented):
Great PR, I just have a few comments but everything looks good. The only doubt I have is about the `recentf-list`.

dashboard.el
Stebalien: I'm preserving the current behavior here. I _believe_ this exists so that if some section (a) modifies the recentf list, some later section (b) won't be affected. E.g., if the recent files section comes after the org agenda section.

dashboard.el
Stebalien: This lambda is currently a closure that reads/modifies the outer state.

dashboard-widgets.el
Stebalien: done

dashboard.el
ricardoricho: Ok, we can leave it that way and maybe later change the recent files behavior in a way that `dashboard-insert-startupify-lists` doesn't know anything about recent files.

I didn't think about each section not been affected, I just thought the "origial" `recentf-list` is saved before any dashboard section consume a file and polluted that list. After all the dashboard is done, we set back the `recentf-list` with  the "origial" list value like nothing happened.

dashboard.el
ricardoricho: Ok, let's leave it as a "local" definition. But my point is that if a closure needs to read from the external state, it is better to have that value as a parameter and try to avoid "side effects" by not modifying the external scope.

Stebalien: I generally agree (and didn't write this code, just moved it).

JesusMtnez:
@ricardoricho @jcs090218 did you have any change to test it?

jcs090218:
I would to like to test it before we merge this! I will probably do it next weekend (hopefully). If @ricardoricho tested and it works, feel free to merge it! ;)

jcs090218 (approved):

seagle0128:
This PR introduced the error: `(void-function image-supported-file-p)`.
Emacs 28.2, on Windows and macOS.

jcs090218:
Ah... okay. I think the function `image-supported-file-p` only exists after `29.1`. @Stebalien Can you make a patch to handle this?

Why are you using the function `dashboard-center-line`? I don't think that's intended to be used externally.

seagle0128:
@jcs090218 I used `dashboard-center-line` to customize some functionalities. It is not a big problem. `dashboard-insert-center` is the alternative. Please fix the issue of `image-supported-file-p`.

PR #409, opened 2022-11-17
Restore Emacs 26 support

Stebalien:
Use `image-type-available-p` and `image-type` instead of `image-supported-file-p` as that function only exists in Emacs 29.


jcs090218 (approved):

jcs090218:
LGTM! Thanks!

@seagle0128 Does this patch works for you?

jcs090218:
Okay, I am merging this since @channel-42 has confirmed it works!

seagle0128:
It works, thank you!

PR #415, opened 2022-12-03
Fix dashboard centering when both an image and text are specified

### LINKED ISSUE #413: The text banner is not centered when using terminal Emacs
Greetings,

When `dashboard-startup-banner` is set to `'logo` and Emacs in opened in the terminal the package defaults to a text banner. This is natural and to be expected. However, that text banner is not centered even though `dashboard-center-content` is set to true. Interestingly, this bug only happens when the package defaults to the text banner but not when the user sets it himself. Here is a picture showing the bug:

![Emacs-dashboard text banner off-center bug](https://user-images.githubusercontent.com/70314018/204420263-27639796-338b-4d54-a0f4-899e39b53f6f.png)

And here is the minimal Emacs configuration needed to replicate my results:

```
; Straight.el setup                                                                                                                                                                                                                           
(defvar bootstrap-version)
(let ((bootstrap-file
       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
      (bootstrap-version 5))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously
         "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
         'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  (load bootstrap-file nil 'nomessage))

; emacs-dashboard minimal setup                                                                                                                                                                                                               
(straight-use-package 'dashboard)

(setq dashboard-startup-banner 'logo)
```

I would have liked to fix the problem myself and submit a fix, but I'm not fluent enough in elisp to do it. I believe the bug was introduced in 48860b18b3b6ffac87b281c417ab11dd0578efc8 since things worked fine before then. This is my first issue I've submitted to a open source project ever, so I apologize in advance for any errors that I may have committed here.

I bid you farewell.

Stebalien:
I made a last minute change here and forgot a dot...

fixes #413

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -620,7 +620,7 @@ Argument IMAGE-PATH path to the image."
                      ;; The quoting is intentional. This is a conditional display spec that will
                      ;; align the banner at redisplay time.
                      `((when (display-graphic-p) . ,image-align-spec)
-                       (when (not (display-graphic-p)) ,text-align-spec)))
+                       (when (not (display-graphic-p)) . ,text-align-spec)))
                     ;; One or the other.
                     (text-width text-align-spec)
                     (image-spec image-align-spec)


jcs090218 (approved):
Thanks you! ❤️

PR #418, opened 2022-12-05
fix: don't assume that PNGs are supported

### LINKED ISSUE #417: Dashboard interrupts startup when using Emacs without X support
I am using the Arch Linux `emacs-nox` package and since the last update dashboard fails to load after inserting the Emacs logo in the dashboard buffer. The logo itself seems to be rendered correctly (see attached picture) but immediately after that Emacs emits `Invalid image type 'png'` and stops evaluating the initfile. When running in daemon mode, Emacs can’t get beyond `(dashboard-setup-startup-hook)` and hangs up indefinitely during evaluation. Disabling the logo completely using `(setq dashboard-startup-banner nil)` seems to do fix the issue, though.
20221204_21h26m44s_grim

Stebalien:
fixes #417

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -529,11 +529,13 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
     (pcase dashboard-startup-banner
       ('nil nil)
       ('official
-       (list :image dashboard-banner-official-png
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
LGTM, thanks! ;)

Issue #419, opened 2022-12-05
Incorrect center text calculation


jcs090218:
I think there is one missing factor causes the miscalculation while centering text. Here are few observations,

If you tried to print out the text in `dashboard-center-text`, the title on the next section got included

```
    package.json  /path/to/project/package.jsonProjects: (p)
```

The text `Projects: (p)` got included in the `line-length` hence the centering will go off when that specific item is the max width.

One trick to solve this issue to make the calculation correct is to add `(insert "\n")` after calling the `item-generator`,

```diff
              (mapc (lambda (els)
                      (let* ((el (or (car-safe els) els))
                             (list-size
                              (or (cdr-safe els)
                                  dashboard-items-default-length))
                             (item-generator
                              (cdr-safe (assoc el dashboard-item-generators))))
                        (push (point) dashboard--section-starts)
                        (funcall item-generator list-size)
                        (goto-char (point-max))
+                       (insert "\n")
                        (when recentf-is-on
                          (setq recentf-list origial-recentf-list))
                        (setq max-line-length
                              (max max-line-length (dashboard-maximum-section-length))))) 
                    dashboard-items)
```

but then you will get an extra newline in the final result.

I don't know what's the best solution, but maybe something to do with the variable `dashboard--section-starts`?

cc @Stebalien

PR #420, opened 2022-12-06
Workaround for incorrect center text calculation 

jcs090218:
For #419, and also some clean up


jcs090218:
@Stebalien I am applying this workaround, feel free to open another PR to improve this!.

PR #436, opened 2023-02-15
Add ascii option to dashboard-startup-banner

### LINKED ISSUE #404: Inline logo
It would be awesome if we didn't need to save custom ASCII logos to a file. Is there a way to configure it so we can pass the ASCII text for the logo itself instead of the file path?

6b6279:
Should close #404.

I have done some rudimentary testing. `(put-text-property (point) (point-max) 'face 'dashboard-text-banner)` seems to not format the text added through `insert` in line 597. Why that’s the case, I unfortunately do not know. Any tips are welcome.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -157,6 +157,11 @@ preserved."
   :type 'string
   :group 'dashboard)
 
+(defcustom dashboard-banner-ascii "EMACS"
+  "String to be shown in place of the startup banner if dashboard-startup-banner is set to 'ascii'."
+  :type 'string
+  :group 'dashboard)
+
 (defcustom dashboard-navigator-buttons nil
   "Specify the navigator buttons.
 The format is: `icon title help action face prefix suffix`.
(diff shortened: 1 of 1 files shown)


dashboard-widgets.el
```diff
@@ -205,12 +210,14 @@ Example:
 (defcustom dashboard-startup-banner 'official
   "Specify the startup banner.
 Default value is `official', it displays the Emacs logo.  `logo' displays Emacs
-alternative logo.  An integer value is the index of text banner.  A string
-value must be a path to a .PNG or .TXT file.  If the value is nil then no banner
-is displayed."
+alternative logo. If set to `ascii', the value of dashboard-banner-ascii
```
jcs090218, line 213: I think it requires at least two spaces after a period.

jcs090218 (commented):
Some nit.

Otherwise, LGTM. 👍

6b6279:
Would that suffice?

Sorry, I had to change the commit message after pushing; I figured the initial message was a bit too open-ended.

jcs090218 (approved):
LGMT! Thanks for your contribution!

andrewbanchich:
I tested this out and noticed that the ASCII dashboard now is horizontally aligned to that the start of the banner is at the center, whereas before it was that the center of the of the banner was at the center.

6b6279:
This is due to the issue I’ve mentioned in my initial comment. `put-text-property` somehow doesn’t affect text inserted in line 597.

andrewbanchich:
Not sure I follow @6b6279 (I'm not great with elisp). I tried adding `(save-excursion (insert txt))` in the `:config` section of my `use-package` and it says the txt symbol's value as variable is void.

6b6279:
> Not sure I follow @6b6279 (I'm not great with elisp). I tried adding `(save-excursion (insert txt))` in the `:config` section of my `use-package` and it says the txt symbol's value as variable is void.

I was talking about the `insert` call in the `dashboard-insert-banner` function. See the pull request that was just merged #440 — sync from upstream and check if the ASCII banner is centered and/or formatted correctly now.

andrewbanchich:
Gotcha, thanks!

jcs090218:
@6b6279 Forgot to mention, would you like to open a PR for documentation?

6b6279:
@jcs090218 Documentation? You mean like a short snippet in the README file on how to use the ASCII banner setting?

jcs090218:
@6b6279 👍

PR #439, opened 2023-02-17
dashboard-open

andyjda:
1) autoloading the `dashboard-refresh-buffer` function so it can be available before the dashboard has been displayed.
This should allow users to decide whether they want to display the dashboard by default at startup. If they don't, they can delay loading the whole package, and still be able to open the dashboard whenever they want

2) renaming `dashboard-refresh-buffer` to `dashboard-open` to better suggest the above use of the function. `dashboard-refresh-buffer` is kept as an alias for consistency and backwards compatibilty.

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -440,11 +440,15 @@ Optional argument ARGS adviced function arguments."
     (when recentf-is-on
       (setq recentf-list origial-recentf-list))))
 
-(defun dashboard-refresh-buffer (&rest _)
-  "Refresh buffer."
+
+;;;###autoload
+(defun dashboard-open (&rest _)
+  "Open (or refresh) the *dashboard* buffer."
   (interactive)
   (let ((dashboard-force-refresh t)) (dashboard-insert-startupify-lists))
   (switch-to-buffer dashboard-buffer-name))
+
+(defalias #'dashboard-refresh-buffer #'dashboard-open)
 
 ;;;###autoload
 (defun dashboard-setup-startup-hook ()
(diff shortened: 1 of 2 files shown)


jcs090218:
Why do we have to rename the function to `dashboard-open`? Would autoload the function be sufficient?

andyjda:
Auto-loading the function would be sufficient in providing this functionality, but I believe the `autoload` changes how the function can be used. Given that, the new name is more descriptive of the function's use and more easily discoverable. "Refresh buffer" makes it sound like a dashboard buffer already exists. "Open dashboard" clarifies that we can use the function to create the dashboard. The refresh behavior is just what happens when we call the function and the buffer already exists.

dashboard.el
```diff
@@ -440,12 +440,16 @@ Optional argument ARGS adviced function arguments."
     (when recentf-is-on
       (setq recentf-list origial-recentf-list))))
 
-(defun dashboard-refresh-buffer (&rest _)
-  "Refresh buffer."
+
```
jcs090218, line 443: Can we remove the extra newline here?

andyjda: Yup, my bad, that shouldn't be there

jcs090218:
You might as well want to update the document

andyjda:
good point, thanks. Just updated the README, let me know if it looks good.

Also, I realized I hadn't linked [this issue](https://github.com/emacs-dashboard/emacs-dashboard/issues/437), which is what made me and another user (@dmgerman) think about this possible edit. Just linking it here to add more context

jcs090218 (approved):

jcs090218:
LGTM. Thank you for your contribution! ;)

PR #440, opened 2023-02-20
Fix spillover bug from #436

6b6279:
See comments in #436.

Due to a discrepancy in point behavior between `insert` and `insert-file-contents`, `insert` must be called in a `save-excursion` block to ensure that the point remains unchanged.


jcs090218 (approved):
Make sense. Thank you! 👍

Issue #405, opened 2022-10-22
Allow customizing the display of tags in agenda


nosewings:
It would be nice to have some control over how/whether tags are displayed in the agenda.

In my case, I just don't want tags to be displayed at all. Making this possible would be as simple as adding a new variable and then changing `dashboard-agenda-entry-format` to account for it.

ricardoricho:
Yes @nosewings, both solutions are good. Maybe with the second provide a couple of custom functions to display or not the tags.

PR #441, opened 2023-02-25
Agenda tags format.

ricardoricho:
Hi, this is for #405.

It defines a custom variable, dashboard-agenda-tags-format, that is pointing to a function that be applied to the tag strings "tag1:tag2:tag3".

It also defines dashboard-agenda--show-tags a function that return the tags, and dashboard-agenda--hide-tags a function that returns nil, so tags are ignored.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -1155,6 +1155,15 @@ each agenda entry."
   :type 'string
   :group 'dashboard)
 
+(defcustom dashboard-agenda-tags-format 'identity
+  "Function to format the org agenda tags.
+Any custom function would receives the tags from `org-get-tags'"
+  :type '(choice
+          (const :tag "Show tags" identity)
+          (const :tag "Hide tags" ignore)
+          (function :tag "Custom function"))
+  :group 'dashboard)
+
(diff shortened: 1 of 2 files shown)


dashboard-widgets.el
```diff
@@ -1207,6 +1216,19 @@ If not height is found on FACE or `dashboard-items-face' use `default'."
...
+  (let ((tags (org-get-tags)))
+    (apply dashboard-agenda-tags-format tags)))
+
+(defun dashboard-agenda--show-tags (&rest tags)
+  "Show all TAGS."
+  tags)
+
+(defun dashboard-agenda--hide-tags (&rest _other_tags)
+  "Ignore TAGS and return nil"
+  nil)
```
jcs090218, line 1230: Maybe we can use `#ignore` function here?  Or make `dashboard-agenda-tags-format` accept `nil`? I think it will be much easier to customize it this way.

ricardoricho: @jcs090218 thanks for the suggestions.

jcs090218 (approved):
LGTM! Let me know when this is ready to merge!

PR #443, opened 2023-03-17
Make the dashboard buffer read-only, as originally intended

vifon:
This code was introduced before lexical binding was in use, so I suspect it used to work. Currently it causes the read-only status of the dashboard buffer to never be set, as only the lexical variable gets set instead.

I believe the original intention was to allow writing into the read-only buffer when generating it. `inhibit-read-only` should be a superior solution.


jcs090218 (approved):

jcs090218:
Thank you!

Issue #444, opened 2023-03-22
Icons on Custom Widgets


ayman:
Is there a way to add an icon to a custom widget? 

`(defun dashboard-insert-heading (heading &optional shortcut)` in `dashboard-widgets.el` has a `cond` which catches the built in section types but I couldn't see how to have that match a custom section with icon.

I have an easy modification to `dashboard-insert-heading` to make this work but wanted to check here first before submitting the pull request.

ricardoricho:
Hi @ayman you're right! That `cond` inside [`dashboard-insert-heading`](https://github.com/emacs-dashboard/emacs-dashboard/blob/b648a45684677aa29cdb00e4d14d02dd9fa9cb68/dashboard-widgets.el#L483) makes the icons insertions closed to other widgets.
Please submit your pull request.

PR #445, opened 2023-03-24
Adding ability and docs for icons in custom widget headers.

ayman:
Addressing #444, icons can be added to custom headers through an optional argument (which leaves the original `dashboard-insert-heading` function compatible and equally operational as before).

Also adds ignore to MacOS temp `.DS_Store` files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 
 # ignore generated files
 *.elc
+.DS_Store
 
 # eask packages
 .eask/
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 
 # ignore generated files
 *.elc
-.DS_Store
 
 # eask packages
 .eask/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -5,6 +5,7 @@
 
 # ignore generated files
 *.elc
+.DS_Store
```
ricardoricho, line 8: I don't think we have to add `.DS_Store` not everybody used a mac.
I believe that if you use one, you could add it to `~/.config/git/ignore` to ignore it from all your projects.

dashboard-widgets.el
```diff
@@ -497,6 +497,8 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
              ((string-equal heading "Projects:")
               (all-the-icons-octicon (cdr (assoc 'projects dashboard-heading-icons))
                                      :height 1.2 :v-adjust 0.0 :face 'dashboard-heading))
+             ((not (null icon))
```
ricardoricho, line 500: This could be one line: `((not (null icon)) icon)`

ricardoricho (changes requested):
Hi @ayman, great work. Having the `icon` as an optional parameter is a nice solution.
I just drop you a few comments.

.gitignore
ayman: Thanks for that tip!

ricardoricho (approved):
Great!

PR #446, opened 2023-03-27
feat(elpaca): provide a semantically correct init time in init info

actondev:
The reported time would be the time needed only to the queue the packages, not to actually process them.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -181,6 +181,8 @@ Example:
       (when (boundp 'straight--profile-cache)
         (setq package-count (+ (hash-table-count straight--profile-cache) package-count)))
       (when (fboundp 'elpaca--queued)
+	(setq time (format "%f seconds" (float-time (time-subtract elpaca-after-init-time
+								   before-init-time))))
         (setq package-count (length (elpaca--queued))))
       (if (zerop package-count)
           (format "Emacs started in %s" time)


jcs090218:
The reported time means after Emacs is fully initialized, so the processing time may be part of the initialized time. It depends on the situation.

Sorry, I don't use elpaca so can't really give any suggestion on this.

@Riyyi implements this in #399, hope he can chime in and help us?

Riyyi:
Elpaca is run asynchronously, so only its queue is started after the init files are done loading. This means that `after-init-time` and `(emacs-init-time)` report an incorrect time. Therefor, Elpaca has an equivalent of this variable `elpaca-after-init-time`, its docs state: `Time after elpaca-after-init-hook is run.` (because the regular `after-init-hook` also wouldn't do). Note that this queue is responsible for package installation but also loading, so is always ran.

I personally have this in my config to make `(emacs-init-time)` report the correct time:
```(add-hook 'elpaca-after-init-hook (lambda () (setq after-init-time (current-time))))```
I tested this commit, and it reports around ~1.2 seconds compared to the above snippet which reports ~1.75 seconds and seems more accurate to reality. I don't know if there is an earlier time variable available, however.

jcs090218:
Ah, okay. Thanks for chiming in and the explanation!

So I assumed this PR is good to merge? Thank you! :)

PR #451, opened 2023-05-02
add nerd-icons support to dashboard

rainstormstudio:
**demo**
(Ubuntu 20.04, Gnome Desktop, Alacritty terminal, FiraCode Nerd Fonts)
image

**Major Changes**
I added `dashboard-icon-type` which specifies which icon package to use (either `'all-the-icons` or `'nerd-icons`). Based on `dashboard-icon-type`, other icon related custom variables such as `dashboard-heading-icons` will have corresponding default values. 

**Configuration for using Nerd Icons**
```elisp
(setq dashboard-display-icons-p t) ;; to use icons under terminal
(setq dashboard-icon-type 'nerd-icons)
```

**Things I am not sure about**
I'm not sure if my implementation for setting default values for some custom variables based on `dashboard-icon-type` is good. Let's say I set `dashboard-icon-type` to `'nerd-icons`. My current implementation for setting the default value of `dashboard-heading-icons` is to use a `pcase` statement. Does setting `dashboard-icon-type` trigger the evaluation of `dashboard-heading-icons`'s `pcase`? If I want to set a new value for `dashboard-heading-icons` in a init file, does it require some kind of ordering like I must setting `dashboard-icon-type` before setting `dashboard-heading-icons`?

**Other comments**
I have been using this dashboard for a few weeks now, and no issues so far. However, it might still have bugs that I did not think of. I'll be tracking nerd icons related issues here in emacs-dashboard, and I will try my best to solve them :)

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -268,36 +268,39 @@ predicate value."
   "Get the formatted icon for DIR.
 ARGS should be a plist containing `:height', `:v-adjust',
 or `:face' properties."
-  (cond ((eq dashboard-icon-type 'all-the-icons) (apply #'all-the-icons-icon-for-dir dir args))
-        ((eq dashboard-icon-type 'nerd-icons) (apply #'nerd-icons-icon-for-dir dir args))))
+  (pcase dashboard-icon-type
+    ('all-the-icons (apply #'all-the-icons-icon-for-dir dir args))
+    ('nerd-icons (apply #'nerd-icons-icon-for-dir dir args))))
 
 (defun dashboard-icon-for-file (file &rest args)
   "Get the formatted icon for FILE.
 ARGS should be a plist containing `:height', `:v-adjust',
 or `:face' properties."
-  (cond ((eq dashboard-icon-type 'all-the-icons) (apply #'all-the-icons-icon-for-file file args))
-        ((eq dashboard-icon-type 'nerd-icons) (apply #'nerd-icons-icon-for-file file args))))
+  (pcase dashboard-icon-type
+      ('all-the-icons (apply #'all-the-icons-icon-for-file file args))
+      ('nerd-icons (apply #'nerd-icons-icon-for-file file args))))
 
 (defun dashboard-octicon (name &rest args)
   "Get the formatted octicon.
 ARGS should be a plist containing `:height', `:v-adjust',
 or `:face' properties."
-  (cond ((eq dashboard-icon-type 'all-the-icons) (apply #'all-the-icons-octicon name args))
-        ((eq dashboard-icon-type 'nerd-icons) (apply #'nerd-icons-octicon name args))))
+  (pcase dashboard-icon-type
+    ('all-the-icons (apply #'all-the-icons-octicon name args))
+    ('nerd-icons (apply #'nerd-icons-octicon name args))))
 
 (defcustom dashboard-footer-icon
   (if (dashboard-display-icons-p)
-      (cond
-       ((eq dashboard-icon-type 'all-the-icons)
-        (all-the-icons-fileicon "emacs"
-                                :height 1.1
-                                :v-adjust -0.05
-                                :face 'font-lock-keyword-face))
-       ((eq dashboard-icon-type 'nerd-icons)
-        (nerd-icons-sucicon "nf-custom-emacs"
-                            :height 1.1
-                            :v-adjust -0.05
-                            :face 'font-lock-keyword-face)))
+      (pcase dashboard-icon-type
+        ('all-the-icons
+         (all-the-icons-fileicon "emacs"
+                                 :height 1.1
+                                 :v-adjust -0.05
+                                 :face 'font-lock-keyword-face))
+        ('nerd-icons
+         (nerd-icons-sucicon "nf-custom-emacs"
+                             :height 1.1
+                             :v-adjust -0.05
+                             :face 'font-lock-keyword-face)))
     (propertize ">" 'face 'dashboard-footer))
   "Footer's icon."
   :type 'string
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -69,6 +69,7 @@
 (defvar org-todo-keywords-1)
 (defvar all-the-icons-dir-icon-alist)
 (defvar package-activated-list)
+(defvar elpaca-after-init-time)
 (declare-function string-pixel-width "subr-x.el")   ; TODO: remove this after 29.1
 (declare-function shr-string-pixel-width "shr.el")  ; TODO: remove this after 29.1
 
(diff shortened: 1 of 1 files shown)


dashboard-widgets.el
```diff
@@ -98,6 +101,53 @@ preserved."
...
+Will be of the form `(list-type . icon-name-string)`.
+If nil it is disabled.  Possible values for list-type are:
+`recents' `bookmarks' `projects' `agenda' `registers'"
+  :type  '(repeat (alist :key-type symbol :value-type string))
+  :group 'dashboard)
+
+(defcustom dashboard-agenda-item-icon
+  (cond
+   ((eq dashboard-icon-type 'all-the-icons) (all-the-icons-octicon "primitive-dot" :height 1.0 :v-adjust 0.01))
+   ((eq dashboard-icon-type 'nerd-icons) (nerd-icons-octicon "nf-oct-primitive_dot" :height 1.0 :v-adjust 0.01)))
```
jcs090218, line 138: To make things consistent, why not use `pcase` everywhere? :D

jcs090218 (commented):
Some nits. Everything else LGTM!

jcs090218:
Can you resolve the compile error? See https://github.com/emacs-dashboard/emacs-dashboard/actions/runs/4864724799/jobs/8675087587?pr=451#step:5:116.

```
In end of data:
dashboard-widgets.el:138:43: Warning: the function `nerd-icons-octicon' is not
    known to be defined.
```

> I'm not sure if my implementation for setting default values for some custom variables based on `dashboard-icon-type` is good. Let's say I set `dashboard-icon-type` to `'nerd-icons`. My current implementation for setting the default value of `dashboard-heading-icons` is to use a `pcase` statement. Does setting `dashboard-icon-type` trigger the evaluation of `dashboard-heading-icons`'s `pcase`?

Yeah, you are right. Users will have to set up the dashboard config early. If you want to change the icon type, they would have to restart Emacs or set them all manually. I don't know if there is a better solution than this, but I don't think this is a huge issue since people normally start the dashboard at the beginning and hope it works. There isn't a lot of runtime tweaking throughout the lifecycle.

> If I want to set a new value for `dashboard-heading-icons` in a init file, does it require some kind of ordering like I must setting `dashboard-icon-type` before setting `dashboard-heading-icons`?

No, the `defcustom` operation won't overwrite variables that are already set.

rainstormstudio:
Thank you so much for the detailed explanation! I just fixed the compile error and replaced those `cond` with `pcase`. Hopefully this is better now :D

dashboard-widgets.el
```diff
@@ -213,14 +264,40 @@ predicate value."
...
+or `:face' properties."
+  (cond ((eq dashboard-icon-type 'all-the-icons) (apply #'all-the-icons-icon-for-file file args))
+        ((eq dashboard-icon-type 'nerd-icons) (apply #'nerd-icons-icon-for-file file args))))
+
+(defun dashboard-octicon (name &rest args)
+  "Get the formatted octicon.
+ARGS should be a plist containing `:height', `:v-adjust',
+or `:face' properties."
+  (cond ((eq dashboard-icon-type 'all-the-icons) (apply #'all-the-icons-octicon name args))
+        ((eq dashboard-icon-type 'nerd-icons) (apply #'nerd-icons-octicon name args))))
```
jcs090218, line 286: And maybe all these places, use `pcase` for consistency. :D

rainstormstudio: oh yeah I should do a search for all of the occurrences. Sorry about that! 

BTW is there any advantage of using `pcase` over `cond`? or it doesn't matter and we just want consistency here?

jcs090218:
> Thank you so much for the detailed explanation! I just fixed the compile error and replaced those `cond` with `pcase`.

No, I should say thank you to you! Thanks for creating nerd-icons! ❤️ It's amazing! I've added all of the nerd-icons related packages to my config! See [here](https://github.com/jcs-emacs/jcs-emacs/blob/87ee0917de783f5ed725bece5c47b70ec357a36c/lisp/jcs-package.el#L290-L293).

dashboard-widgets.el
jcs090218: It's a matter of taste. I prefer `pcase` since the code is cleaner. The downside of this, it will add up startup time due to loading the `pcase.el` module.

jcs090218: It's a matter of taste. I prefer `pcase` since the code is cleaner. The downside of this, it will add up startup time due to loading the `pcase.el` module.

rainstormstudio:
I just replaced other `cond` with `pcase`. Please take a look.

Yeah I felt jealous when I saw Vim users had Nerd icons and we didn't. That's why I made nerd-icons lol

dashboard-widgets.el
```diff
@@ -98,6 +102,53 @@ preserved."
   :type 'integer
   :group 'dashboard)
 
+(defcustom dashboard-icon-type (or (require 'nerd-icons nil t)
+                                   (require 'all-the-icons nil t))
```
jcs090218, line 106: I think it's good to prevent require in the root level. Let's prevent it by making sure the `dashboard-set-heading-icons` is non-nil:

```elisp
(and dashboard-set-heading-icons
    (or (require 'nerd-icons nil t)
        (require 'all-the-icons nil t)))
```

Make sure to place `dashboard-set-heading-icons` to the front to avoid compile error.

dashboard-widgets.el
```diff
@@ -98,6 +102,53 @@ preserved."
...
+  "Icon type used for dashboard.
+The value can be one of: `all-the-icons', `nerd-icons'."
+  :type 'symbol
+  :group 'dashboard
+  :set
+  (lambda (k v)
+    (and (eq v 'all-the-icons) (not (require 'all-the-icons nil t)) (setq v nil))
+    (and (eq v 'nerd-icons) (not (require 'nerd-icons nil t)) (setq v nil))
```
jcs090218, line 114: I think we can use `pcase` here top?

```elisp
(pcase v
  ('all-the-icons
   (unless (require 'all-the-icons nil t)
     (setq v nil)))
  ('nerd-icons
   (unless (require 'nerd-icons nil t)
     (setq v nil))))
```

jcs090218 (commented):
Some clean up! ;)

jcs090218 (approved):
Some work needs to be done on our side. Anyways, the PR is good to be merged! :D

Thank you for your contribution! ❤️ 🚀

rainstormstudio:
Sorry I was a bit busy just now.

The clean up looks good, and I think we also need to modify the README regarding using nerd-icons.

Thank you for your guidance!

jcs090218:
> Sorry I was a bit busy just now.

No worries! The code quality is there, therefore it's good to merge! :D 

> The clean up looks good, and I think we also need to modify the README regarding using nerd-icons.

Can you open a PR for this. :)

PR #463, opened 2023-05-12
fix: dashboard not appearing on startup

### LINKED ISSUE #433: Dashboard not appearing at startup after updating package
Hello!

I've updated to the most recent version of `emacs-dashboard` and now the dashboard no longer appears when I start emacs. Prior to updating the package, my emacs would start and display the dashboard. I use emacs 28.1 if that matters.

When emacs starts, a dashboard buffer appears but it is an empty buffer. Calling `M-x dashboard-refresh-buffer` displays the dashboard as normal.

I saw https://github.com/emacs-dashboard/emacs-dashboard/issues/421. The issue seems the same, however I do not use chemacs.

amartos:
The commit partially revert 48860b1 by re-adding the dashboard-resize-on-hook function and its setup hook.

The setup hook has also been moved into the dashboard-setup-startup-hook function.

Fixes #433.

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -453,12 +453,25 @@ Optional argument ARGS adviced function arguments."
 
 (defalias #'dashboard-refresh-buffer #'dashboard-open)
 
+(defun dashboard-resize-on-hook (&optional _)
+  "Re-render dashboard on window size change."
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Thank you!

Issue #459, opened 2023-05-08
nil icon name when dashboard-set-heading-icons turned on


sunng87:
When `dashboard-set-heading-icons` turned on, I'm getting error: `unable to find icon with name nil in icon set` with both nerd-icons and all-the-icons. Not sure if it's because additional headings in dashboard-widgets like ls-directory.

jcs090218:
What's the value of your variable `dashboard-icon-type`? It should either be ``all-the-icons` or `'nerd-icons`.

sunng87:
I've tested with both all-the-icons and nerd-icons, the error persists until I turned off heading icons and keeps only file icons. So I believe it has something to do with heading icon set.

mikaelspringer:
I experience the same thing. When I set `dashboard-set-heading-icons` to `t`, I get the error message and no icons are shown in `*dashboard*`. When I set `dashboard-set-heading-icons` to `nil`, the error message disappear and icons are shown again in *dashboard*.

One other strange thing is that when I do `M-x customize-group dashboard` a lot of the icon related variables are marked `CHANGED outside Customize. (mismatch)`. I know for sure I haven't changed them.

jcs090218:
What's the value of variable `dashboard-heading-icons`? It seems like it's nil and it shouldn't be. 🤔 

It could be an issue regarding the order you load this package, it should be something similar to:

```elisp
(setq dashboard-set-heading-icons t)
(require 'dashboard)  ; it will auto configure it for you
```

or set them all manually:

```elisp
(require 'dashboard)
(setq dashboard-set-heading-icons t)
(setq dashboard-icon-type 'all-the-icons)
(setq dashboard-heading-icons '((recents   . "history")
                                (bookmarks . "bookmark")
                                (agenda    . "calendar")
                                (projects  . "rocket")
                                (registers . "database")))
```

sunng87:
After checking `customize-variable`, I found `dashboard-heading-icons` is set to `nil`.

knaveightt:
Hello everyone - will also raise that I am experiencing the same issue. After reviewing this thread, I also tried to set the icons manually as well, but to no avail. Here is the exact config I am using, utilizing use-package for reference. Willing to try alternative approaches based on feedback:

```emacs-lisp
  (use-package dashboard
    :config
    (dashboard-setup-startup-hook)
    (setq dashboard-items '((recents  . 5)
			    (bookmarks . 5)
			    (projects . 5)
[...]

PR #467, opened 2023-07-02
Fixes issue #459: Display icons for headings and files using Nerd fonts

xenobis:
Hi,

this pull request fixes issue #459 using Nerd fonts. There are changes to several defuns in the nerd-icons package, which are replaced by macros, apparently the return values for the icons changed as well. Did not analyze it in full detail.

My setup:
1. Windows 11 22H2 (Build 22621.1848)
2. Emacs 28.2
3. nerd-icons-20230609.641
4. dashboard-20230512.1839

Before fix:

grafik

After fix:

grafik

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -192,7 +192,8 @@ If nil it is disabled.  Possible values for list-type are:
 (defcustom dashboard-agenda-item-icon
   (pcase dashboard-icon-type
     ('all-the-icons (all-the-icons-octicon "primitive-dot" :height 1.0 :v-adjust 0.01))
-    ('nerd-icons (nerd-icons-octicon "nf-oct-dot_fill" :height 1.0 :v-adjust 0.01)))
+    ;; MRO 20230625 change to TODO-Item org: nf-oct-dot_fill -> nf-seti-todo
+    ('nerd-icons (nerd-icons-sucicon "nf-seti-todo" :height 1.0 :v-adjust 0.01)))
   "Agenda item icon."
   :type 'string
   :group 'dashboard)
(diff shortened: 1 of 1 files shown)


dashboard-widgets.el
```diff
@@ -302,26 +303,23 @@ Optional argument REP is the replacement string of non-displayable character."
   "Get the formatted icon for DIR.
 ARGS should be a plist containing `:height', `:v-adjust',
 or `:face' properties."
-  (dashboard-replace-displayable
    (pcase dashboard-icon-type
      ('all-the-icons (apply #'all-the-icons-icon-for-dir dir args))
-     ('nerd-icons (apply #'nerd-icons-icon-for-dir dir args)))))
+     ('nerd-icons (nerd-icons-icon-for-dir dir args))))
```
jcs090218, line 308: Why are we removing `apply` here? 🤔

jcs090218:
The idea of the function `dashboard-replace-displayable` is to replace characters (to empty string) that cannot be displayed on the screen. 🤔

dashboard-widgets.el
xenobis: As far as I know `apply` does not work with macros. In my setup it wasn't working.

xenobis:
> The idea of the function `dashboard-replace-displayable` is to replace characters (to empty string) that cannot be displayed on the screen. 🤔

OK, but it has currently the effect to be "empty" when it is used with the return values for the icons, so no icons are displayed.

jcs090218:
Can you try debug it with the function `char-displayable-p`? It should return `t` if the character is displayable.

seagle0128:
The root cause is `dashboard-replace-displayable`.

jcs090218:
The issue has been documented in FAQ.org; therefore, I am closing this.

Issue #408, opened 2022-11-08
Bookmarks section causing TRAMP to open remote connections


jdm204:
My emacs version: `GNU Emacs 29.0.50 (build 3, x86_64-pc-linux-gnu, GTK+ Version 3.24.33, cairo version 1.16.0) of 2022-11-05`

During startup, enabling the bookmarks section causes TRAMP to open a remote to connection for bookmarks which have a remote path, and so prompts for a password etc.

When I run `bookmark-all-names` and `bookmark-get-filename` for a remote bookmark it doesn't seem to open a connection, so it seems that it shouldn't be necessary to do so to start up the dashboard.

jeslie0:
I am having the same issue with remote projects also.

jdm204:
Thanks for looking into this @ricardoricho, I am confused because I have `recentf-auto-cleanup` as `mode`, but it's not an issue with the bookmarks commented out of my dashboard config:

```
dashboard-items '((recents  . 10)
                       ;;(bookmarks . 10)
                       (agenda . 10)
                       (projects . 10)
                       (registers . 10))
```
if I uncomment bookmarks, I have the issue.

ricardoricho:
I'm sorry @jdm204 but  I couldn't reproduce the bug. I created two bookmarks, one with sudo and other with ssh, restarted Emacs and everything is good, it didn't ask for any password.

@jeslie0 which Emacs version do you have?

I'm in Emacs 28.2 for mac.  Maybe it has something to do with a newer version.

> I'm not proficient at elisp so apologies in advance but is there any chance `bookmark-jump` could get called without the user doing anything like [here](https://github.com/emacs-dashboard/emacs-dashboard/blob/554dc6fac1362dd6b66318c8250eea8bd63aa92f/dashboard-widgets.el#L955)?

That lambda (`bookmark-jump`) is only called when user hit enter over the item (bookmark)

jeslie0:
I'm using Emacs 29.0.50. I make a minimal working example and see if it fails there, however I have done the following tests to see when it works:

- When using a /sudo: based bookmark, I am not prompted for a password when starting Emacs (but I am to start the sudo connection),
- When using a /ssh:user@localhost, I am not prompted for a password when starting Emacs (but I am on the initial ssh connection),
- When using /ssh:remote-machine: I am prompted for a password when starting Emacs and when creating initial the ssh connection.

jeslie0:
[...]

PR #480, opened 2023-10-10
fix: Make remove entry optional

jcs090218:
For #408.


lckarssen:
Thanks! :pray: That works for me :+1:

Issue #332, opened 2021-10-18
Add a face for `dashboard-footer-icon`


cyberknight777:
By adding this, it would make customizing the footer-icon better.

o0nd7ots:
+1

jcs090218:
I somehow overlooked this feature. Anyways, feel free to open PRs for this feature! :)

jcs090218:
The face `dashboard-footer-icon-face` is added. You can do:

```elisp
(setq dashboard-footer-icon (nerd-icons-faicon "nf-fa-heart" :face 'dashboard-footer-icon-face))
```

o0nd7ots:
This is great. Thanks ricardricho!

PR #483, opened 2023-10-30
Add dashboard icon face.

ricardoricho:
This is for #332

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -251,12 +251,6 @@ Example:
   :type '(function string)
   :group 'dashboard)
 
-(defcustom dashboard-footer
-  (nth (random (1- (1+ (length dashboard-footer-messages)))) dashboard-footer-messages)
-  "A footer with some short message."
-  :type 'string
-  :group 'dashboard)
-
 (defcustom dashboard-display-icons-p #'display-graphic-p
   "Predicate to determine whether dashboard should show icons.
 Can be nil to not show icons and any truthy value to show them.  When set to a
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):
LGTM! Thank you!

dashboard-widgets.el
```diff
@@ -251,12 +251,6 @@ Example:
   :type '(function string)
   :group 'dashboard)
 
-(defcustom dashboard-footer
-  (nth (random (1- (1+ (length dashboard-footer-messages)))) dashboard-footer-messages)
-  "A footer with some short message."
-  :type 'string
-  :group 'dashboard)
```
jcs090218, line 258: Why are we removing the random footer? 🤔

jcs090218 (commented):
I've got a closer look. I think you have accidentally removed `dashboard-footer` (not the face).

dashboard-widgets.el
ricardoricho: We are not using that custom variable. The `dashboard-insert-footer` function uses `dashboard-random-footer` to get the value.  The `dashboard-footer-message` list is used to customize the text, from the README:
```emacs-lisp
(setq dashboard-footer-messages '("Dashboard is pretty cool!"))
```

jcs090218: Good catch! Now sure why we still have this variable in the first place. 🤔

jcs090218:
Thank you! Merged!

jcs090218:
Just to let you know, I have reverted the `propertize`  in `dashboard-insert-footer` function (22786237e16cfeae33f07ae9c5eeaf061408579a) since propertize will make the icon undisplayable. 🤔 

Before changes:



After changes:

PR #491, opened 2024-01-27
don't open dashboard if inhibit-startup-screen is enabled

timcharper:
if some other process has set inhibit-startup-screen before the dashboard has shown, respect it and do not display dashboard

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -464,19 +464,20 @@ Optional argument ARGS adviced function arguments."
 
 (defun dashboard-initialize ()
   "Switch to dashboard and run `dashboard-after-initialize-hook'."
-  (switch-to-buffer dashboard-buffer-name)
-  (goto-char (point-min))
-  (redisplay)
-  (run-hooks 'dashboard-after-initialize-hook))
+  (when (and (< (length command-line-args) 2) ;; Assume no file name passed
+             (not inhibit-startup-screen)) ;; Don't display if something already set inhibit-startup-screen
+    (switch-to-buffer dashboard-buffer-name)
+    (goto-char (point-min))
+    (redisplay)
+    (run-hooks 'dashboard-after-initialize-hook)))
 
 ;;;###autoload
 (defun dashboard-setup-startup-hook ()
   "Setup post initialization hooks unless a command line argument is provided."
-  (when (< (length command-line-args) 2) ;; Assume no file name passed
-    (add-hook 'window-size-change-functions #'dashboard-resize-on-hook 100)
-    (add-hook 'window-setup-hook #'dashboard-resize-on-hook)
-    (add-hook 'after-init-hook #'dashboard-insert-startupify-lists)
-    (add-hook 'emacs-startup-hook #'dashboard-initialize)))
+  (add-hook 'window-size-change-functions #'dashboard-resize-on-hook 100)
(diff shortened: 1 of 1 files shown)


timcharper:
I'm wondering if there is a way to better integrate the dashboard with the existing startup.el code. There's already logic in there to not display the startup screen in certain conditions, and the intent of dashboard.el is usually to replace the startup screen (it would be awful to have both the startup screen _AND_ emacs dashboard show at the same time)

dashboard.el
```diff
@@ -464,19 +464,20 @@ Optional argument ARGS adviced function arguments."
...
 
 ;;;###autoload
 (defun dashboard-setup-startup-hook ()
   "Setup post initialization hooks unless a command line argument is provided."
-  (when (< (length command-line-args) 2) ;; Assume no file name passed
-    (add-hook 'window-size-change-functions #'dashboard-resize-on-hook 100)
-    (add-hook 'window-setup-hook #'dashboard-resize-on-hook)
-    (add-hook 'after-init-hook #'dashboard-insert-startupify-lists)
-    (add-hook 'emacs-startup-hook #'dashboard-initialize)))
+  (add-hook 'window-size-change-functions #'dashboard-resize-on-hook 100)
```
timcharper, line 477: I'm wondering why these aren't added any time dashboard-open is invoked

jcs090218: You need to add this before `dashboard-open` in order for some features to work. For example, see the variable `dashboard-init-info`.

dashboard.el
```diff
@@ -464,19 +464,20 @@ Optional argument ARGS adviced function arguments."
 
 (defun dashboard-initialize ()
   "Switch to dashboard and run `dashboard-after-initialize-hook'."
-  (switch-to-buffer dashboard-buffer-name)
-  (goto-char (point-min))
-  (redisplay)
-  (run-hooks 'dashboard-after-initialize-hook))
+  (when (and (< (length command-line-args) 2) ;; Assume no file name passed
+             (not inhibit-startup-screen)) ;; Don't display if something already set inhibit-startup-screen
```
jcs090218, line 468: Why did you move those if conditions here? It should be taken care of in the function `dashboard-setup-startup-hook`.

jcs090218:
> I'm wondering if there is a way to better integrate the dashboard with the existing startup.el code. There's already logic in there to not display the startup screen in certain conditions, and the intent of dashboard.el is usually to replace the startup screen (it would be awful to have both the startup screen AND emacs dashboard show at the same time)

My understanding is that the dashboard intends to replace the startup screen but not the `startup.el` (it may sound contradicting). `inhibit-startup-screen` should only be used in a `startup.el`, not the dashboard. We try to make dashboard snappier, but it would take some decent work to replace `startup.el`.

There are also historical reasons: the dashboard was created in Emacs 25.1, and `startup.el` has significantly improved after all these years. It's even more difficult since we are still supporting 26.1 and above. I honestly don't think that's a good idea. 🤔 I don't want to be too assertive here, so feel free to share your thoughts.

timcharper:
Fair enough, I trust your judgement here :)

I had to put the check for inhibit-startup-screen to make it work with my automation, and just moved the other condition with it.

However, after using it, I'm noticing dashboard doesn't open at all at all :( 

I'll just close this PR for now. If I come up with a better solution, I'll suggest it.

PR #493, opened 2024-01-30
Add option to vertically center dashboard

cvdub:
This PR adds an option to vertically center the dashboard content. I tested it with a few different dashboard configurations on MacOS and everything seems to be working.

Before:
Screenshot 2024-01-29 at 4 54 07 PM

After:
Screenshot 2024-01-29 at 4 54 23 PM

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -445,10 +445,12 @@ Optional argument ARGS adviced function arguments."
         (dashboard-insert-footer)
         (when dashboard-vertically-center-content
           (goto-char (point-min))
-          (let* ((buffer-lines (count-lines (point-min) (point-max)))
-                 (vertical-padding (floor (/ (- (window-height) buffer-lines) 2))))
-            (when (> vertical-padding 0)
-              (insert (make-string vertical-padding ?\n)))))
+          (when-let* ((content-height (cdr (window-absolute-pixel-position (point-max))))
+                      (vertical-padding (floor (/ (- (window-pixel-height) content-height) 2)))
+                      ((> vertical-padding 0))
+                      (vertical-lines (/ vertical-padding
+                                         (line-pixel-height))))
+            (insert (make-string vertical-lines ?\n))))
         (goto-char (point-min))
         (dashboard-mode)))
     (when recentf-is-on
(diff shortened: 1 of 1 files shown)


dashboard.el
```diff
@@ -438,6 +443,12 @@ Optional argument ARGS adviced function arguments."
           (delete-char -1)
           (insert dashboard-page-separator))
         (dashboard-insert-footer)
+        (when dashboard-vertically-center-content
+          (goto-char (point-min))
+          (let* ((buffer-lines (count-lines (point-min) (point-max)))
+                 (vertical-padding (floor (/ (- (window-height) buffer-lines) 2))))
+            (when (> vertical-padding 0)
+              (insert (make-string vertical-padding ?\n)))))
```
jcs090218, line 451: This won't work since the line height isn't an accurate calculation. Try calculating it in pixels, see below:

```elisp
        (when dashboard-vertically-center-content
          (goto-char (point-min))
          (when-let* ((content-height (cdr (window-absolute-pixel-position (point-max))))
                      (vertical-padding (floor (/ (- (window-pixel-height) content-height) 2)))
                      ((> vertical-padding 0))
                      (vertical-lines (/ vertical-padding (line-pixel-height))))
            (insert (make-string vertical-lines ?\n))))
```

cvdub: Thanks for the suggestion!

dashboard.el
```diff
@@ -438,6 +443,14 @@ Optional argument ARGS adviced function arguments."
           (delete-char -1)
           (insert dashboard-page-separator))
         (dashboard-insert-footer)
+        (when dashboard-vertically-center-content
+          (goto-char (point-min))
+          (when-let* ((content-height (cdr (window-absolute-pixel-position (point-max))))
+                      (vertical-padding (floor (/ (- (window-pixel-height) content-height) 2)))
+                      ((> vertical-padding 0))
+                      (vertical-lines (/ vertical-padding
+                                         (line-pixel-height))))
```
jcs090218, line 452: Is there a reason why we want to put an extra newline here? 🤔

jcs090218 (approved):

Issue #481, opened 2023-10-23
Transparency issue with PNGs


o0nd7ots:
This problem arose as colatteral of issue #479.

I want to have an image with transparency as my dashboard banner.
At first I started with `.svg` format and later, then advised, I switched to `.png`.
SVG hasn't got a background set and when I export it to PNG and open with Okular, it indeed is transparent.

Do any of you know what might be causing this?

jcs090218:
FYI, I have seen this issue. But I am currently busy. I will try to get back to you as soon as I have time! 😓

o0nd7ots:
It's good to know that you are aware of it. I decied that it is better to separate two problems into two issues.

Also this is not something urgent, so there's no need to hurry. Thank you for your time though!

A7R7:
This is a bug of emacs, not dashboard, as it's described in https://mail.gnu.org/archive/html/bug-gnu-emacs/2022-12/msg01080.html
If you want a TRUELY transparent dashboard banner, try using XPM format image.

o0nd7ots:
Okay, this is lame XD.

I checked it and `.webp` also works.
I wish this gets fixed by the emacs team.

It is not connected to the dashboard itself, so I close the issue

PR #495, opened 2024-02-12
add `dashboard-image-extra-props' option.

dwuggh:
This should be helpful to #481, I can display transparent images with setting it to '(:mask heuristic).


jcs090218 (approved):
Thank you!

Issue #501, opened 2024-03-04
Animated webp images are not rendering like gifs do.


SATYADAHAL:
Gifs are automatically rendered when in emacs banner. But webp images are not rendered. Webp images can also have multiple frames like gif and can be animated.
Animated webp [example](https://mathiasbynens.be/demo/animated-webp).
Moreover webp images in emacs works well with transparent background. Many other format like gif,png doesnot support transparency in emacs but webp does. 
### Dashboard banner with gif.(Rendered but background not transparent)
![gif_output](https://github.com/emacs-dashboard/emacs-dashboard/assets/51309409/1a7a18b0-ee16-4e65-9030-eb505d6af5f7)

### Dashboard banner with webp.(Background trasparent but not rendered) 
![webp_output](https://github.com/emacs-dashboard/emacs-dashboard/assets/51309409/4e5ecbb5-0186-477c-853b-237e4183360c)

Emacs can render the webp image by itself. So shouldn't this be possible to render webp image in dashboard?

PR #502, opened 2024-03-04
Fixed: Adding anmiation to webp image.

SATYADAHAL:
I think i got around this #501. I am a beginner at lisp. I just took help from google and LLM and fixed the issue for me. Check if its a good approach to this problem.
### Working animated webp.
webp_working_output

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -712,19 +712,13 @@ Argument IMAGE-PATH path to the image."
 
 
 (defun dashboard--is-image-animated-p (image-path)
-  "Return t if animated and has multiple frames, nil otherwise.
-   String -> bool.
-   Argument IMAGE-PATH: Path to the image.
-   This function first checks if the image type supports animation 
-   like GIF or WebP, and then verifies if the supported image has 
-   multiple frames."
-  (when (or (eq (image-type image-path) 'gif)
-            (eq (image-type image-path) 'webp))
-    (let ((image (create-image image-path)))
-      (if (image-multi-frame-p image)
-          t
-        nil))))
-
+  "Return if image has multiple frames.
+String -> bool.
+Argument IMAGE-PATH path to the image."
+   (when (or (eq (image-type image-path) 'gif)
+           (eq (image-type image-path) 'webp))
+   (let ((image (create-image image-path)))
+     (image-multi-frame-p image))))
 
 (defun dashboard--type-is-xbm-p (image-path)
   "Return if image is a xbm.
@@ -778,9 +772,7 @@ Argument IMAGE-PATH path to the image."
                                          (memq 'scale (funcall 'image-transforms-p)))
                                 img-props))))))
         (add-text-properties start (point) `(display ,image-spec))
-        
-(when (dashboard--is-image-animated-p img)
-  (image-animate image-spec 0 t)))
+        (when (dashboard--is-image-animated-p img) (image-animate image-spec 0 t)))
 
       
       ;; Finally, center the banner (if any).
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -710,6 +710,12 @@ String -> bool.
 Argument IMAGE-PATH path to the image."
   (eq 'gif (image-type image-path)))
 
+(defun dashboard--type-is-webp-p (image-path)
+  "Return t if the image at IMAGE-PATH is a WebP image, nil otherwise.
+   String -> bool.
+   Argument IMAGE-PATH path to the image."
+  (eq 'webp (image-type image-path)))
+
 (defun dashboard--type-is-xbm-p (image-path)
   "Return if image is a xbm.
 String -> bool.
@@ -750,6 +756,8 @@ Argument IMAGE-PATH path to the image."
           (setq image-spec
                 (cond ((dashboard--type-is-gif-p img)
                        (create-image img))
+                      ((dashboard--type-is-webp-p img)
+                       (create-image img))
                       ((dashboard--type-is-xbm-p img)
                        (create-image img))
                       ((image-type-available-p 'imagemagick)
@@ -760,7 +768,11 @@ Argument IMAGE-PATH path to the image."
                                          (memq 'scale (funcall 'image-transforms-p)))
                                 img-props))))))
         (add-text-properties start (point) `(display ,image-spec))
-        (when (dashboard--type-is-gif-p img) (image-animate image-spec 0 t)))
+        
+(when (or (dashboard--type-is-gif-p img) (dashboard--type-is-webp-p img))
+  (image-animate image-spec 0 t)))
+
+      
       ;; Finally, center the banner (if any).
       (when-let* ((text-align-spec `(space . (:align-to (- center ,(/ text-width 2)))))
                   (image-align-spec `(space . (:align-to (- center (0.5 . ,image-spec)))))
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -710,11 +710,21 @@ String -> bool.
 Argument IMAGE-PATH path to the image."
   (eq 'gif (image-type image-path)))
 
-(defun dashboard--type-is-webp-p (image-path)
-  "Return t if the image at IMAGE-PATH is a WebP image, nil otherwise.
+
+(defun dashboard--is-image-animated-p (image-path)
+  "Return t if animated and has multiple frames, nil otherwise.
    String -> bool.
-   Argument IMAGE-PATH path to the image."
-  (eq 'webp (image-type image-path)))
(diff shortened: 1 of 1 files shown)


dashboard-widgets.el
```diff
@@ -750,6 +756,8 @@ Argument IMAGE-PATH path to the image."
           (setq image-spec
                 (cond ((dashboard--type-is-gif-p img)
                        (create-image img))
+                      ((dashboard--type-is-webp-p img)
```
ricardoricho, line 759: This "cond" would change to `(dashboard--image-animated-p img) (create-image img)` for the `gif` and the `webp` cases.

dashboard-widgets.el
```diff
@@ -710,6 +710,12 @@ String -> bool.
 Argument IMAGE-PATH path to the image."
   (eq 'gif (image-type image-path)))
 
+(defun dashboard--type-is-webp-p (image-path)
```
ricardoricho, line 713: Since `webp` images and `gif` images are going to be animated. I suggest an `--image-animated-p` that tells if the image is a `gif` or a `webp`.

ricardoricho (changes requested):

dashboard-widgets.el
SATYADAHAL: I was digging much into the images formats. Then i found out that gifs and webp can have single frame too.
So we don't need to animate those gifs and wepb with single frames.
I will try to create a function that checks if the image has multiple frame.
But if my function will return true only if image is (gif or webp) and  image has multiple frame.
I will make some changes. Tell me what do you think about it.

dashboard-widgets.el
SATYADAHAL: @ricardoricho i have implemeted a function to check if image has multiple frame. What do you think about this solution?

dashboard-widgets.el
```diff
@@ -710,6 +710,22 @@ String -> bool.
...
   (eq 'gif (image-type image-path)))
 
+
+(defun dashboard--is-image-animated-p (image-path)
+  "Return t if animated and has multiple frames, nil otherwise.
+   String -> bool.
+   Argument IMAGE-PATH: Path to the image.
+   This function first checks if the image type supports animation 
+   like GIF or WebP, and then verifies if the supported image has 
+   multiple frames."
```
jcs090218, line 720: Make sure to keep the style consistent.

dashboard-widgets.el
```diff
@@ -710,6 +710,22 @@ String -> bool.
...
+  (when (or (eq (image-type image-path) 'gif)
+            (eq (image-type image-path) 'webp))
+    (let ((image (create-image image-path)))
+      (if (image-multi-frame-p image)
+          t
+        nil))))
```
jcs090218, line 726: There is no reason to return `nil`. If you return nothing, it will default to `nil`. Or even better, return the condition directly:

```elisp
  (when (or (eq (image-type image-path) 'gif)
            (eq (image-type image-path) 'webp))
    (let ((image (create-image image-path)))
      (image-multi-frame-p image)))
```

dashboard-widgets.el
```diff
@@ -760,7 +778,11 @@ Argument IMAGE-PATH path to the image."
                                          (memq 'scale (funcall 'image-transforms-p)))
                                 img-props))))))
         (add-text-properties start (point) `(display ,image-spec))
-        (when (dashboard--type-is-gif-p img) (image-animate image-spec 0 t)))
+        
+(when (dashboard--is-image-animated-p img)
+  (image-animate image-spec 0 t)))
+
+
```
jcs090218, line 785: The indentation seems a bit off. Make sure there are no extra newlines. 🤔

jcs090218 (commented):
Some nits.

dashboard-widgets.el
SATYADAHAL: I also though of this one initially. But I did this one just to return boolean(nil or t). The function image-multi-frame-p will return   a cons (nimages . delay), where nimages is the number of frames and delay is the delay in seconds between them. I think cons (nimages . delay) is considered true. So your fix  must be good to go with.

dashboard-widgets.el
SATYADAHAL: @jcs090218 i have made the changes.

ricardoricho (changes requested):
Great work @SidharthArya there are small details to fix and there is the `(nimages . delay)` "validation" of a "heavy" image, but that's another issue.

dashboard-widgets.el
```diff
@@ -710,6 +710,16 @@ String -> bool.
...
   (eq 'gif (image-type image-path)))
 
+
+(defun dashboard--is-image-animated-p (image-path)
+  "Return if image has multiple frames.
+String -> bool.
+Argument IMAGE-PATH path to the image."
+   (when (or (eq (image-type image-path) 'gif)
+           (eq (image-type image-path) 'webp))
+   (let ((image (create-image image-path)))
```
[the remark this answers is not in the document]
SATYADAHAL: @ricardoricho @jcs090218 I have now made changes so that the function only checks image type(gif or webp) and multi-frame is checked before animating on this [line](https://github.com/SATYADAHAL/emacs-dashboard/blob/93a70def782a98a47d1d39fea990f9dec8043bf0/dashboard-widgets.el#L771).

dashboard-widgets.el
```diff
@@ -748,7 +758,9 @@ Argument IMAGE-PATH path to the image."
                          (list :max-height dashboard-image-banner-max-height))
                        dashboard-image-extra-props)))
           (setq image-spec
-                (cond ((dashboard--type-is-gif-p img)
+                (cond  ((dashboard--is-image-animated-p img)
+                       (create-image img))
+                      ((dashboard--type-is-gif-p img)
```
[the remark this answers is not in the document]
SATYADAHAL: Done removing

dashboard-widgets.el
SATYADAHAL: fixed

ricardoricho (approved):
Great!

dashboard-widgets.el
```diff
@@ -704,11 +704,12 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
...
 
-(defun dashboard--type-is-gif-p (image-path)
-  "Return if image is a gif.
+(defun dashboard--image-animated-p (image-path)
+  "Return if image is a gif or webp.
 String -> bool.
 Argument IMAGE-PATH path to the image."
-  (eq 'gif (image-type image-path)))
+  (or (eq 'gif (image-type image-path))
+      (eq 'webp (image-type image-path))))
```
jcs090218, line 712: Can we use `memq` here?

dashboard-widgets.el
```diff
@@ -748,7 +749,7 @@ Argument IMAGE-PATH path to the image."
                          (list :max-height dashboard-image-banner-max-height))
                        dashboard-image-extra-props)))
           (setq image-spec
-                (cond ((dashboard--type-is-gif-p img)
+                (cond  ((dashboard--image-animated-p img)
```
jcs090218, line 752: There is one extra space here. 🤔

dashboard-widgets.el
```diff
@@ -760,7 +761,9 @@ Argument IMAGE-PATH path to the image."
                                          (memq 'scale (funcall 'image-transforms-p)))
                                 img-props))))))
         (add-text-properties start (point) `(display ,image-spec))
-        (when (dashboard--type-is-gif-p img) (image-animate image-spec 0 t)))
+        (when  (image-multi-frame-p image-spec) (image-animate image-spec 0 t)))
```
jcs090218, line 764: One extra space after `when`.

jcs090218 (commented):
More nits.

dashboard-widgets.el
SATYADAHAL: I will fix it

jcs090218 (approved):

jcs090218:
Thank you! :)

dashboard-widgets.el
```diff
@@ -760,7 +760,9 @@ Argument IMAGE-PATH path to the image."
                                          (memq 'scale (funcall 'image-transforms-p)))
                                 img-props))))))
         (add-text-properties start (point) `(display ,image-spec))
-        (when (dashboard--type-is-gif-p img) (image-animate image-spec 0 t)))
+        (when (image-multi-frame-p image-spec) (image-animate image-spec 0 t)))
```
basil-conto: Maybe this doesn't make a difference in practice, but FWIW `image-multi-frame-p` can be non-`nil` for non-animated images too, such as multi-frame TIFFs. It's the `cdr` of `image-multi-frame-p` that distinguishes a multi-frame from an animated image.

PR #510, opened 2024-03-17
Support for random banners.

DevelopmentCool2449:
Hello, this patch allow to users display random banners in a list.
I'd like to hear their reviews.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -347,9 +347,7 @@ ARGS should be a plist containing `:height', `:v-adjust', or `:face' properties.
 
 (defcustom dashboard-startup-banner 'official
   "Specify the banner type to use.
-
 Value can be
- - nil display no banner
  - 'official  displays the official Emacs logo.
  - 'logo  displays an alternative Emacs logo.
  - an integer which displays one of the text banners.
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -357,8 +357,7 @@ Value can be
    supported files types are gif/image/text/xbm
  - a cons of 2 strings which specifies the path of an image to use
    and other path of a text file to use if image isn't supported.
- - a list that can display an random banner, must contain 2 or more
-   arguments, supported values are:
+ - a list that can display an random banner, supported values are:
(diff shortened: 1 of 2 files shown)


jcs090218:
One warning:

```
dashboard-widgets.el:348:2: Warning: custom-declare-variable ‘dashboard-startup-banner’ docstring has wrong usage of unescaped single quotes (use \= or different quoting)
```

dashboard-widgets.el
```diff
@@ -740,7 +758,7 @@ Argument IMAGE-PATH path to the image."
       (when graphic-mode (insert "\n"))
       ;; If specified, insert a text banner.
       (when-let ((txt (plist-get banner :text)))
-        (if (eq dashboard-startup-banner 'ascii)
+        (if (length= banner 2)
```
jcs090218, line 761: This looks weird. Try avoid the magic number here. 🤔

DevelopmentCool2449:
> One warning:
> 
> ```
> dashboard-widgets.el:348:2: Warning: custom-declare-variable ‘dashboard-startup-banner’ docstring has wrong usage of unescaped single quotes (use \= or different quoting)
> ```

Fixed.

dashboard-widgets.el
DevelopmentCool2449: > This looks weird. Try avoid the magic number here. 🤔

That was the only way i found to check if banner is a cons cell, using listp or consp conflicts with random banner list.

DevelopmentCool2449: If you know a way to determine a cons cell, i'd really appreciate if you could tell me.

DevelopmentCool2449: Fixed, only determining if it is a list fixs this issue.

jcs090218:
It seems like it doesn't work inside the terminal. 🤔

DevelopmentCool2449:
> It seems like it doesn't work inside the terminal. 🤔 

Fixed

jcs090218 (approved):

jcs090218:
Thank you!

PR #511, opened 2024-03-18
Doom style support

DevelopmentCool2449:
This is an implementation to support doom dashboard like widgets,
I added only a few of basic widgets for this.
I'd like to hear their reviews and comments about this idea :-).

Here the final result:
screenshot_2


jcs090218:
Supporting a specific style is not our intention, and it's beyond this project's scope. 🤔 Creating a separate package might be a better option.

DevelopmentCool2449:
> Supporting a specific style is not our intention, and it's beyond this project's scope. 🤔 Creating a separate package might be a better option.

I think you are right, i will split this PR because there are other improvements that i would like pull.

PR #512, opened 2024-03-18
Add new hook for dashboard.

DevelopmentCool2449:
I've noticed that using dashboard-after-initialize-hook with solaire doesn't fix banner image background issue.
I found that this is because the banner is created before solaire is called.

The solution i found is making other hook that just run before dashboard buffer is created.

I wonder if this is a can a good idea or just run dashboard-after-initialize-hook in dashboard--with-buffer.

Also i've noticed moving solaire first in dashboard-startupify-list fix this too.

### DIFF
--- a/dashboard.el
+++ b/dashboard.el
@@ -81,13 +81,13 @@
     map)
   "Keymap for dashboard mode.")
 
-(defcustom dashboard-after-initialize-hook nil
-  "Hook that is run after dashboard buffer is initialized."
+(defcustom dashboard-before-initialize-hook nil
+  "Hook that is run before dashboard buffer is initialized."
   :group 'dashboard
   :type 'hook)
 
-(defcustom dashboard-before-initialize-hook nil
-  "Hook that is run before dashboard buffer is initialized."
+(defcustom dashboard-after-initialize-hook nil
+  "Hook that is run after dashboard buffer is initialized."
   :group 'dashboard
   :type 'hook)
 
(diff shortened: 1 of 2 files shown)


dashboard.el
```diff
@@ -440,6 +445,7 @@ Optional argument ARGS adviced function arguments."
   "Execute BODY in dashboard buffer."
   (declare (indent 0))
   `(with-current-buffer (get-buffer-create dashboard-buffer-name)
+     (run-hooks 'dashboard-before-initialize-hook)
```
jcs090218, line 448: I don't think this should be inserted here.

dashboard.el
```diff
@@ -86,6 +86,11 @@
   :group 'dashboard
   :type 'hook)
 
+(defcustom dashboard-before-initialize-hook nil
+  "Hook that is run before dashboard buffer is initialized."
+  :group 'dashboard
+  :type 'hook)
```
jcs090218, line 92: Let’s put this above the after-initialize hook, so it makes a bit more sense.

DevelopmentCool2449: fixed

dashboard.el
DevelopmentCool2449: should i move this into dashboard-insert-startupify-lists?

this hook must be run before dashboard inserts any widgets.

DevelopmentCool2449:
I've moved before-initialize hook in insert-startupify-lists.

dashboard.el
jcs090218: Yes.

DevelopmentCool2449:
I forgot delete dashboard-before-initialize-hook from dashboard--with-buffer.
I've fixed it.

jcs090218 (approved):

jcs090218:
Thank you

PR #513, opened 2024-03-19
Support for arguments in dashboard-startupify-list and add FAQ

DevelopmentCool2449:
This patch adds optional support for insert arguments into dashboard-startupify-list, this allow insert amount newlines in dashboard-insert-newline, and add into README how to hide modeline.

Also this updates dashboard-items for optionally insert list size.

### DIFF
--- a/README.org
+++ b/README.org
@@ -276,16 +276,7 @@ To use it with [[https://github.com/hlissner/emacs-hide-mode-line][hide-mode-lin
 
 Or for doom-modeline users
 #+begin_src elisp
-  (setq doom-modeline-mode-alist
-     (delete (assoc 'dashboard-mode doom-modeline-mode-alist)
-             doom-modeline-mode-alist))
-  ;; Or for use-package
-  (use-package doom-modeline
-   :custom
-   (doom-modeline-mode-alist
-    (delete (assoc 'dashboard-mode doom-modeline-mode-alist)
-             doom-modeline-mode-alist))
-   ...)
+  (add-to-list 'doom-modeline-mode-alist '(dashboard-mode))
 #+end_src
 
 To use it with [[https://github.com/ericdanan/counsel-projectile][counsel-projectile]] or [[https://github.com/bbatsov/persp-projectile][persp-projectile]]
--- a/README.org
+++ b/README.org
@@ -167,32 +167,6 @@ To modify the widget heading name:
                                ("Agenda for the coming week:" . "Agenda:")))
 #+END_SRC
 
-The default length for each item.
-#+BEGIN_SRC elisp
-  (setq dashboard-items-default-length 20)
-#+END_SRC
-
-To add your own custom widget is pretty easy, define your widget's callback function and add it to `dashboard-items` as such:
-#+BEGIN_SRC elisp
-  (defun dashboard-insert-custom (list-size)
-    (insert "Custom text"))
-  (add-to-list 'dashboard-item-generators  '(custom . dashboard-insert-custom))
-  (add-to-list 'dashboard-items '(custom) t)
-#+END_SRC
-
-To add an icon to a custom widget, insert it with `dashboard-insert-heading` in your custom function.  In this example, there is an icon but no shortcut.
-#+BEGIN_SRC elisp
-  (defun dashboard-insert-custom (list-size)
-    (dashboard-insert-heading "News:"
-                              nil
-                              (all-the-icons-faicon "newspaper-o"
-                                                    :height 1.2
-                                                    :v-adjust 0.0
-                                                    :face 'dashboard-heading))
-    (insert "\n")
-    (insert "    Custom text"))
-#+END_SRC
-
 To use ~all-the-icons~ package:
 #+BEGIN_SRC emacs-lisp
   (setq dashboard-icon-type 'all-the-icons)  ; use `all-the-icons' package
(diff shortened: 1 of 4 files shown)


dashboard.el
```diff
@@ -500,8 +505,13 @@ See `dashboard-item-generators' for all items available."
...
 
-        (mapc (lambda (fn)
-                (funcall fn))
+        (mapc (lambda (list)
+                (if-let* (((consp list))
+                          (fn (car list))
+                          (args (cdr list))
+                          ((not (eq fn 'lambda))))
+                    (funcall fn args)
+                  (funcall list)))
```
jcs090218, line 514: I'm a bit confused. Can you give me an example on how to use this?

README.org
```diff
@@ -265,6 +265,20 @@ Or for use-package users
...
+  ;; Or for use-package
+  (use-package hide-mode-line
+    :hook (dashboard-mode . hide-mode-line-mode)
+    ...)
+#+end_src
+
+Or for doom-modeline users
+#+begin_src elisp
+  (add-to-list 'doom-modeline-mode-alist '(dashboard-mode))
+#+end_src
```
jcs090218, line 280: It's a good time to place all these in the FAQ section. 🤔 

Maybe a FAQ.org.

dashboard.el
DevelopmentCool2449: Sure,
For example i want to add 3 or 4 spaces to startupify,
the only way to do this is adding dashboard-insert-newline
multiple times, this make the list longer and repetitive.

```
(setq dashboard-startupify-list 
                             '(dashboard-insert-banner
                               dashboard-insert-newline
                               dashboard-insert-newline
                               dashboard-insert-newline
                               dashboard-insert-newline
                               ...)
```

Using a cons cell avoid making this multiple times:
```
(setq dashboard-startupify-list 
                             '(dashboard-insert-banner
                               (dashboard-insert-newline . 4)
                               ...)
```
Even this works with other functions
like for example deleting 1 or 2 spaces/newlines:
```
(setq dashboard-startupify-list 
                             '(dashboard-insert-banner
                               (delete-char . -4)
                               ...)
```

>                           ((not (eq fn 'lambda))))

I'd added a conditional to determine if It's not a lambda because it throw error.

README.org
DevelopmentCool2449: Added

dashboard.el
ricardoricho: Hi, I have the following setup:
```emacs-lisp
(dashboard-startupify-list `(dashboard-insert-banner
                                 dashboard-insert-newline
                                 dashboard-insert-banner-title
                                 dashboard-insert-newline
                                 dashboard-insert-init-info
                                 ,(dashboard-insert-newline 2)
                                 dashboard-insert-navigator
                                 dashboard-insert-newline
                                 dashboard-insert-items))
```
And I change `dashboard-insert-newline` as follows:
```emacs-lisp
(defun dashboard-insert-newline (&optional times)
  "When called without an argument, insert a newline.
When called with TIMES return a function that insert TIMES number of newlines."
  (if times
      (lambda ()
        (insert (make-string times (string-to-char "\n") t)))
    (insert "\n")))
```
I think we can have this "special" dashboard functions that can be used in the list.
What do you think? Or do we need to support any function?

DevelopmentCool2449: > ```emacs-lisp
> (defun dashboard-insert-newline (&optional times)
>   "When called without an argument, insert a newline.
> When called with TIMES return a function that insert TIMES number of newlines."
>   (if times
>       (lambda ()
>         (insert (make-string times (string-to-char "\n") t)))
>     (insert "\n")))
> ```
> 
> I think we can have this "special" dashboard functions that can be used in the list. What do you think?

I like this idea, however using that setup throws error for me: `insert: Symbol’s value as variable is void: times`

ricardoricho: Ok, then, I don't think we need the `if-let` part just `apply` the function, something like this:
```emacs-lisp
(lambda (entry)
  (if (listp entry) (apply (car entry) (cdr entry))
    (funcall entry)))
```
What was the error using `lambda`?

DevelopmentCool2449: > Ok, then, I don't think we need the if-let part just apply the function, something like this:
> ```
> (lambda (entry)
>   (if (listp entry) (apply (car entry) (cdr entry))
>     (funcall entry)))
> ```

Sure, it can work.

> What was the error using lambda?

I'm not sure if using lambda inside a defun and later backquote it can compromise perfomance.
I'm updating the patch for include all of this, but i'm not sure about lambda.

ricardoricho: Oh, great, don't worry about the lambdas, with this "entry apply" approach there's no need for the newline definition to return a lambda, it could be just the insert part, like you have before.

DevelopmentCool2449: > Oh, great, don't worry about the lambdas, with this "entry apply" approach there's no need for the newline definition to return a lambda, it could be just the insert part, like you have before.

Good to hear!, i was worrying that those lambdas could make dashboard slower because i heard something like that on old post.

For now, i don't think there can be other stuffs to add to this PR.

jcs090218:
I've tried your branch but it doesn't seem to work. 🤔 

```elisp
(setq dashboard-startupify-list
      `(dashboard-insert-banner
        ,(dashboard-insert-newline 3)
        dashboard-insert-banner-title
        dashboard-insert-newline
        dashboard-insert-init-info
        dashboard-insert-items
        dashboard-insert-newline
        dashboard-insert-footer))
```

DevelopmentCool2449:
> I've tried your branch but it doesn't seem to work. 🤔
> 
> ```emacs-lisp
> (setq dashboard-startupify-list
>       `(dashboard-insert-banner
>         ,(dashboard-insert-newline 3)
>         dashboard-insert-banner-title
>         dashboard-insert-newline
>         dashboard-insert-init-info
>         dashboard-insert-items
>         dashboard-insert-newline
>         dashboard-insert-footer))
> ```

What error did you get?

DevelopmentCool2449:
I think i've found the issue, now it is fixed (i hope)

jcs090218 (approved):

jcs090218:
LGTM. Thank you! :)

PR #519, opened 2024-03-26
Support for random footer icon

DevelopmentCool2449:
This play well and makes sense with `random-footer-messages`


jcs090218 (approved):

jcs090218:
Thank you!

PR #521, opened 2024-03-26
Fix #471

### LINKED ISSUE #471: No icons in dashboard on Windows
I am using Emacs 28.2/29.1 on Windows 11, with the latest dashboard package. There is no icon for files and sections on Windows, but all are fine on macOS/Ubuntu/WSL, even in terminals. I tried emacs -Q, and the issue still there.

My configurations are [here](https://github.com/seagle0128/.emacs.d/blob/master/lisp/init-dashboard.el).

The screenshot is below.

image

DevelopmentCool2449:
`dashboard-replace-displayable` doesn't return `nerd-icon` string due apparently `internal-char-font` only return non-nil if it is only displayable with current frame font, 
this is a problem to some user that doesn't using any nerd patched font as default, also due nerd-font can use any "unused" char-set for its icons it's necessary return `string` if nerd-icons is avalaible. 
The same applies to all-the-icons

This may close: https://github.com/emacs-dashboard/emacs-dashboard/issues/471 and https://github.com/emacs-dashboard/emacs-dashboard/pull/467


DevelopmentCool2449:
I'm closing this due to low activity.

jcs090218:
Sorry, I'm a bit busy and still need to investigate the issue. 😓

PR #525, opened 2024-04-07
Add new entry to FAQ.org

DevelopmentCool2449:
Due Windows port doesn't handle well with icons in dashboard
(see https://github.com/emacs-dashboard/emacs-dashboard/issues/471), I think it's good idea add a new entry 
into FAQ file about How to fix this issue.

I've also improve it a little.


jcs090218 (approved):

jcs090218:
Thanks for taking care of this! :D

PR #532, opened 2024-05-29
Fix custom type for dashboard-heading-icons

skissue:
The previous type is wrong, as it's for a list of alists, rather than a single alist.


jcs090218 (approved):

jcs090218:
Thank you!

PR #537, opened 2024-08-11
Allow propertized ascii banner text

DevelopmentCool2449:
This a simple change for allow to the user colorize/propertize the banner ascii _ad_ _libitum_, such as a rainbow banner and more.

If the banner text is propertized, it will not apply dashboard-text-banner face.

I've not tested it, but since it's an only small change it should work properly


jcs090218 (approved):

jcs090218:
LGTM! Thanks!

PR #541, opened 2024-09-09
feat: `dashboard-banner-logo-symbol-png`

gs-101:
logo-symbol

Dashboard banner of the current Emacs logo, save the background and shadow.

It was made by exporting the [SVG](https://git.savannah.gnu.org/cgit/emacs.git/tree/etc/images/icons/hicolor/scalable/apps/emacs.svg) with Inkscape, using the following options:
- No interlacing
- Compression: 1
- PHYs DPI: 600
- Antialias: 3


jcs090218:
I don't think we should add another logo, or else there will be too many logos in this package. Users can customize it themselves. 🤔

gs-101:
Fair enough, I'll close this PR for now.

jcs090218:
Thank you for your understanding!

PR #546, opened 2024-09-24
feat: Parameters for file icon height and v-adjust

christoffer-arvidsson:
This PR adds support for tweaking the height and v-adjust of file icons. It aligns with the existing parameters
`dashboard-heading-icon-height` and `dashboard-heading-icon-v-adjust` which work similarly for heading icons.

Defaults are set to the previous defaults.

Let me know if there are any changes needed. :slightly_smiling_face:

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -903,7 +903,7 @@ Argument IMAGE-PATH path to the image."
                                           (not (string-equal icon ""))
                                           (not (string-equal title "")))
                                  (propertize " " 'face `(:inherit (variable-pitch
-                                                                  ,face))))
+                                                                   ,face))))
                                (when title (propertize title 'face face)))
                          :help-echo help
                          :action action
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -207,12 +207,12 @@ If nil it is disabled.  Possible values for list-type are:
   :type 'float
   :group 'dashboard)
 
-(defcustom dashboard-file-icon-height 1.0
+(defcustom dashboard-icon-file-height 1.0
   "The height of the file icons."
   :type 'float
   :group 'dashboard)
 
-(defcustom dashboard-file-icon-v-adjust -0.05
+(defcustom dashboard-icon-file-v-adjust -0.05
   "The v-adjust of the file icons."
   :type 'float
   :group 'dashboard)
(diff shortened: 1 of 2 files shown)


dashboard-widgets.el
```diff
@@ -207,6 +207,16 @@ If nil it is disabled.  Possible values for list-type are:
   :type 'float
   :group 'dashboard)
 
+(defcustom dashboard-file-icon-height 1.0
```
ricardoricho, line 210: I suggest `dashboard-icon-file-height` because is a custom of `dashboard-icon` family ¿?

dashboard-widgets.el
```diff
@@ -207,6 +207,16 @@ If nil it is disabled.  Possible values for list-type are:
...
+  "The height of the file icons."
+  :type 'float
+  :group 'dashboard)
+
+(defcustom dashboard-file-icon-v-adjust -0.05
```
ricardoricho, line 215: My suggestion is `dashboard-icon-file-v-adjust`

ricardoricho (changes requested):
Thanks for your PR. There is also a `dashboard-icon-for-file`. Should we have `dashboard-icon-dir-height` and `dashboard-icon-dir-v-adjust`, or could we use only one `dashboard-icon-height` and `dashboard-icon-v-adjust` for both?

christoffer-arvidsson:
> Thanks for your PR. There is also a `dashboard-icon-for-file`. Should we have `dashboard-icon-dir-height` and `dashboard-icon-dir-v-adjust`, or could we use only one `dashboard-icon-height` and `dashboard-icon-v-adjust` for both?

I think that's sensible. One issue might be that the there's already different v-adjusts for the different icons used `-0.05` for file, `0.01` for others. It looked fine for me to set these to the same with the exception of the footer which could get the same treatment as heading if needed.

It could be nice to try to organize this a bit better before someone starts customizing them?

jcs090218:
I think it makes sense just to have the `-file` variables since there is only `dashboard-set-file-icons` and no `dashboard-set-dir-icons`.

jcs090218:
Can you update the REAMDE as well? Thanks!

christoffer-arvidsson:
@jcs090218 good catch. dir/remote is now handled too. I also updated the README as you requested (hopefully in a good place).

jcs090218 (approved):

jcs090218:
Thank you for taking care of this!  🚀

PR #547, opened 2024-10-01
Fix vertically-center.

ricardoricho:
Dashboard-open use dashboard--with-buffer and switch to it before inserting the startupify list.

Remove dashboard-force-refresh and pass it as optional argument to force-refresh the buffer.


jcs090218 (approved):

jcs090218:
Thank you!

PR #570, opened 2025-03-29
Fix #566 to show "installed packages" instead of "loaded packages".

### LINKED ISSUE #566: Isn't the claim "X packages loaded" missleading?
Just had an argument regarding this.

Image

If you defer loading with use-package, for example, are all those packages loaded?

Shouldn't we say "X packages installed, Emacs loaded in Y seconds" (or some variation of that) instead?

ricardoricho:
Also, change defcustom to defun in `dashboard-init-info`, this remove the possibility to customize a text when `dashboard-insert-init-info` but there is `dashboard-banner-logo-title` which is more suitable for that purpose.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -890,9 +890,13 @@ Argument IMAGE-PATH path to the image."
 ;;; Initialize info
 (defun dashboard-insert-init-info ()
   "Insert init info."
-  (let ((init-info (if (functionp dashboard-init-info)
-                       (funcall dashboard-init-info)
-                     dashboard-init-info)))
+  (let ((init-info (cond ((stringp dashboard-init-info)
+                          dashboard-init-info)
+                         ((functionp dashboard-init-info)
+                          (funcall dashboard-init-info))
+                         (t
+                          (user-error "Unknown init info type (%s): %s"
+                                      (type-of init-info) init-info)))))
     (dashboard-insert-center
      (propertize init-info 'face 'font-lock-comment-face))))
 
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -288,7 +288,12 @@ Example:
                                        (const nil)))))
   :group 'dashboard)
 
-(defun dashboard-init-info ()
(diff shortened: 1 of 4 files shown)


jcs090218:
I would still prefer to have the custom variable `dashboard-init-info`. There isn't a reason to remove it. 🤔

ricardoricho:
Hi, I restored the custom variable, but with a twist. Now, the user can set the variable to a function that receives two arguments, the `init-time` (for example, "2.1 seconds.") and the `package-count`, for example:

```emacs-lisp
(defun my-custom-init-info (init-time package-count)
  "This is my custom init function that uses INIT-TIME and PACKAGE-COUNT."
  (format "Emacs installed %d packages in %s" package-count init-time)))
```
After `(setq dashboard-init-info 'my-custom-init-info)` the `dashboard-init-info` will produce: `Emacs installed 76 packages in 1.23 seconds.`

If someone want a custom text, could ignore the arguments with `&rest`:

```emacs-lisp
(setq dashboard-init-info (lambda (&rest _) "My custom init message"))
```

What do you think? Is it confusing?

dashboard-widgets.el
```diff
@@ -876,9 +886,9 @@ Argument IMAGE-PATH path to the image."
 ;;; Initialize info
 (defun dashboard-insert-init-info ()
   "Insert init info."
-  (let ((init-info (if (functionp dashboard-init-info)
-                       (funcall dashboard-init-info)
-                     dashboard-init-info)))
+  (let* ((init-time (dashboard-init--time))
+         (packages-count (dashboard-init--packages-count))
+         (init-info (funcall dashboard-init-info init-time packages-count)))
```
jcs090218, line 891: I think the variable `dashboard-init-info` accepts a string as well. 🤔 See https://github.com/emacs-dashboard/emacs-dashboard/blob/master/docs/FAQ.org#how-can-i-change-my-init-message.

jcs090218 (approved):

jcs090218:
I assumed is good to merge! :)

Merged! Thank you!

PR #584, opened 2025-08-18
new logos for no-window mode

simonLeary42:
I think the default emacs logo in no-window mode is ugly.
```
######## ##     ##    ###     ######   ######
##       ###   ###   ## ##   ##    ## ##    ##
##       #### ####  ##   ##  ##       ##
######   ## ### ## ##     ## ##        ######
##       ##     ## ######### ##             ##
##       ##     ## ##     ## ##    ## ##    ##
######## ##     ## ##     ##  ######   ######
```

I made some transparent-background ANSI-art emacs logos with the help of [TerminalImageViewer](https://github.com/stefanhaustein/TerminalImageViewer). I did this by just removing the background color, or removing the foreground color and adding the inverse/reverse escape code.

<table>
<tr>
<td />
<td>my normal background</td>
<td>red background (proof of transparency)</td>
</tr>
<tr>
<td>8 bit color</td>
<td>
image
</td>
<td>
image
</td>
</tr>
<tr>
<td>true color</td>
<td>
image
</td>
<td>
image
</td>
</tr>
</table>

The screenshots above were taken in iTerm2. It looks fine in [Windows Console](https://en.wikipedia.org/wiki/Windows_Console), Windows Terminal, Kitty.

[Terminal](https://en.wikipedia.org/wiki/Terminal_(macOS)) doesn't support true color, and it displays ascii block characters poorly:

image

I suspect that the pink around the borders can be fixed. [TIV README](https://github.com/stefanhaustein/TerminalImageViewer#common-problems--troubleshooting) says:

> If you see strange horizontal lines, the characters don't fully fill the character cell. Remove additional line spacing in your terminal app.

But I don't see any such setting for Terminal.

mobaxterm has a problem similar to Terminal where it has vertical stripes instead of horizontal. [hterm](https://chromium.googlesource.com/apps/libapps/+/HEAD/hterm/) has horizontal and vertical lines.

GNOME Terminal doesn't like a few of the characters, but I bet that can be fixed:

image

I don't have access to a GUI linux machine with sudo privileges right now to install a bunch of other terminal emulators and take screenshots.

I also did braille:

<table>
<tr><td>iTerm2</td><td>Terminal</td></tr>
<tr>
<td>
image
</td>
<td>
image
</td>
</tr>
</table>

Note that the artifacts in the lower right are just my shell prompt and cursor on the following line.

After creating these logos, I added them as options for `dashboard-startup-banner`, and made it work properly by injecting `ansi-color-apply-on-region` early.

Since `'logo` automatically switches between image and text logos based on device support, I thought it was natural to do the same for true color / 256 color using `display-color-cells`. But since it looks bad on some terminal emulators, I decided against that.

I changed the default value for the logo to braille. I think that's a big improvement, but it only works if we are confident that 99% of users have a font installed with braille characters.

### DIFF
--- a/dashboard-widgets.el
+++ b/dashboard-widgets.el
@@ -778,11 +778,11 @@ When called with TIMES return a function that insert TIMES number of newlines."
                (list :image dashboard-banner-logo-png))
              (list :text dashboard-banner-logo-braille)))
     ('logo-ansi-truecolor
-     (append (when (and (not (display-graphic-p)) (>= (display-color-cells) (expt 2 24)))
+     (append (when (>= (display-color-cells) (expt 2 24))
                (list :text dashboard-banner-logo-ansi-truecolor))
              (list :text dashboard-banner-logo-braille)))
     ('logo-ansi-256color
-     (append (when (and (not (display-graphic-p)) (>= (display-color-cells) 256))
+     (append (when (>= (display-color-cells) 256)
                (list :text dashboard-banner-logo-ansi-256color))
              (list :text dashboard-banner-logo-braille)))
     ('logo-braille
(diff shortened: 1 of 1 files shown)


simonLeary42:
I am not attached to the naming convention, feel free to change it as you see fit

PR #602, opened 2026-04-10
Recents refactor

ricardoricho:
This is a refactor of the "recents" section. It changes the way we keep a copy of `recentf-list` before we visit all the files involved in the construction of the different sections and then restore the list to remove all the pollution of the construction.

One part I'm not sure of is that if `recentf-mode` is not enable, do we have to
enable it? In the current implementation we are activating the mode to read the
list, but there is `recetf-load-list` that do that without activating the mode.

I added some test just for this section, but I expect that this will be the
pattern, to have test files for each section and test some features instead of
testing each function. What do you think?


test/activate.el
```diff
@@ -21,11 +21,11 @@
...
-
-(setq initial-buffer-choice (lambda () (get-buffer-create "*dashboard*")))
-
-(message "Done activation test.")
+(ert-deftest activate-dashboard-with-dashboard-open-test ()
+  (use-package dashboard
+    :load-path "./"
+    :demand t)
+  (with-current-buffer (dashboard-open)
+    (should (string-match-p "Welcome to Emacs!" (buffer-string)))))
```
jcs090218: We would need to manually install `use-package` for early Emacs versions.

Makefile
```diff
@@ -21,9 +21,6 @@ test:
 	@echo "Testing..."
 	$(EASK) test ert ./test/*.el
 
-test-activate: package install
-	$(EASK) emacs --batch -l ./test/activate.el
```
jcs090218, line 25: Might just replace this with ert tests?

PR #610, opened 2026-07-25
Fix shortcut setup invoking the cycling command during render

### LINKED ISSUE #609: dashboard-insert-shortcut runs the cycling command while the dashboard is drawn ("Only one tabable widget")
Since #597 every redraw of the dashboard buffer prints "Only one tabable widget" in the echo area (with `dashboard-items` starting with a section that has entries, e.g. projects).

I tracked it down to `dashboard-insert-shortcut`:

```elisp
(eval-after-load 'dashboard
  (dashboard--define-shorcut-key-binding ,shortcut-id ,shortcut-char))
```

The second argument of `eval-after-load` is not quoted, so it is evaluated on the spot and `eval-after-load` receives the return value of `dashboard--define-shorcut-key-binding`. That return value is the last `define-key`, i.e. the backward cycling command. And since the `dashboard` feature is already loaded by the time sections are inserted, `eval-after-load` funcalls that command immediately. In other words, inserting a section also runs `widget-backward` in the middle of building the buffer. While the first section is being inserted the buffer only has one widget, so `widget-move` prints the message.

Backtrace captured from an instrumented `message`:

```
  message("Only one tabable widget")
  widget-move(-1)
  widget-backward(1)
  #f(compiled-function () (interactive nil) #<bytecode -0xbe1244998f6c685>)()
  eval-after-load(dashboard #f(compiled-function () (interactive nil) #<bytecode -0xbe1244998f6c685>))
  dashboard-insert-projects(4)
  ...
  dashboard-insert-items()
  dashboard-insert-startupify-lists()
  run-hooks(after-init-hook delayed-warnings-hook)
```

The shortcut keys still get bound correctly, because the bindings happen as side effects before the value is returned. The visible symptoms are the message and point being moved around during the redraw.

Since `dashboard` is by definition loaded when this code runs, the `eval-after-load` wrapper does not seem needed at all. Calling the function directly fixes it for me:

```elisp
(dashboard--define-shorcut-key-binding ,shortcut-id ,shortcut-char)
```

Quoting the form would work too if you want to keep `eval-after-load`.

Seen with dashboard 20260402.436 from MELPA on Emacs 30.2. The code is unchanged on current master.

marktran:
Fixes #609.

`dashboard-insert-shortcut` expands to

```elisp
(eval-after-load 'dashboard
  (dashboard--define-shorcut-key-binding ,shortcut-id ,shortcut-char))
```

The second argument is not quoted, so `eval-after-load` gets the return value of `dashboard--define-shorcut-key-binding`, which is the backward cycling command from the last `define-key`. Since the `dashboard` feature is already loaded when sections are inserted, `eval-after-load` funcalls that command right away. The result is `widget-backward` running in the middle of building the dashboard buffer, which moves point around and echoes "Only one tabable widget" while the first section is the only widget so far.

The feature is always loaded by the time this expansion runs, so the wrapper is unnecessary. This just calls the function directly. The key bindings are unchanged, they were already happening as a side effect.

Tested by byte compiling both files and confirming the shortcut keys are still bound after a render, with no message and no point movement.


test/test-shortcuts.el
```diff
@@ -0,0 +1,49 @@
...
+;;
+;; ERT tests for section shortcut setup in dashboard.
+;;
+
+;;; Code:
+
+(require 'ert)
+(require 'dashboard)
+
+(ert-deftest dashboard-insert-shortcut-binds-cycle-keys ()
```
ricardoricho: The convention I'm using is to name the file `dashboard-shortcuts-tests.el`.
Also, in test definitions I'm used to name them package-function-test, so this could be: 
`dashboard-shorcuts-insert-test` but since, we are not in the "dashboard-shortcuts" package (yet, it would be nice to have one) maybe rename it to `dashboard-insert-shortcut-test`.

test/test-shortcuts.el
```diff
@@ -0,0 +1,49 @@
...
+  "Setting up a section shortcut binds cycling commands in `dashboard-mode-map'."
+  (with-temp-buffer
+    (insert "Recent Files:\n")
+    (dashboard-insert-shortcut 'recents "r" "Recent Files:"))
+  (should (commandp (lookup-key dashboard-mode-map "r")))
+  (should (commandp (lookup-key dashboard-mode-map "R"))))
+
+(ert-deftest dashboard-insert-shortcut-does-not-run-commands ()
```
ricardoricho: It's hard to test that "something" does not call other commands. The should
assertions are about positions (points), maybe the name of this test could be
something about the positions o navigations, beside the `(point-min) assertion
we could test after calling the shortcut command the (point) is in the right
position.

ricardoricho (changes requested):
Thanks

test/test-shortcuts.el
marktran: Renamed the file to `dashboard-shortcuts-tests.el` and the binding test to `dashboard-insert-shortcut-test`.

test/test-shortcuts.el
marktran: Renamed this to `dashboard-insert-shortcut-navigation-test`. It now creates two widgets, verifies setup preserves point, invokes the bound shortcut command, and checks that point moves to the first item.

ricardoricho (approved):
Great, thanks a lot

